PR #198, opened 2026-05-06
feat: encode filter state into URL for bookmarkable and shareable filtered views

### LINKED ISSUE #74: Feature: Encode active filters in the URL so filtered views can be bookmarked and shared
### Summary
Right now, if a user filters by *Python + Security* and wants to share that view with a friend, they have to tell them manually which filters to select. Encoding the filter state into the URL (e.g., `?lang=python&cat=security&q=fuzzing`) would let users **bookmark and share** any filtered view instantly.

### What needs to be done
1. When any filter/search changes, update the browser URL using `history.replaceState()` with query params:
   - `q` → search text
   - `cat` → selected category
   - `lang` → selected language(s), comma-separated
   - `sort` → sort order
2. On page load, read these params with `URLSearchParams` and pre-apply the filters.
3. The page title/URL should update without a full reload.

### Files likely involved
- `app.js` — Read/write URL params alongside filter state changes

### Skills required
- Basic JavaScript
- Familiarity with `URLSearchParams` and `history.replaceState()`

### Helpful resources
- [MDN: URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
- [MDN: history.replaceState](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState)

---
> 💡 A super useful feature — and a great way to learn about the browser History API!

Harshith1702:
Closes #74

## What I did
- On every filter/search change, URL is updated with `history.replaceState()` using `q`, `cat`, `lang`, `sort` query params
- On page load, URL params are read with `URLSearchParams` and pre-applied to filters before rendering
- No full page reload, URL updates silently in the background

No other changes to logic or UI.

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -555,6 +555,14 @@ function applyFilters(){
   renderGrid(res);
   document.getElementById('countDisplay').textContent=res.length;
   document.getElementById('filteredStat').textContent=res.length;
+
+  // Sync filter state to URL
+  const params = new URLSearchParams();
+  if (search)    params.set('q',search);
+  if (cat)    params.set('cat',cat);
+  if (lang)    params.set('lang',lang);
+  if (sort && sort !== 'alpha')    params.set('sort',sort);
+  history.replaceState(null,'',params.toString()?'?'+params.toString():location.pathname);
 }
 
 // Umbrella orgs: link goes to org page, not a single example repo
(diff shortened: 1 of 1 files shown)


Harshith1702:
hey, fixed the CodeRabbit flagged issue in the 2nd commit, pill selections are now persisted as comma separated values in the URL and restored on page load. CodeRabbit will re review once the rate limit clears !!!

S3DFX-CYBER (approved):

PR #550, opened 2026-05-14
updated ideas page url, mentors and channels for  the org

zish-s:


S3DFX-CYBER:
@zish-s can u pls use the templates made for gssoc and also sign the commits for dco

PR #932, opened 2026-05-18
Fixed Actively Maintained filter chip

rishabhx29:
1. 
applyFilters() — Added the missing else if (activeChip === 'active') branch at line 2706. It checks o._gh?.activity === 'active' when GitHub data has been fetched, and gracefully shows all orgs when no data is available yet (!o._gh || fallback).
2. 
fetchStats() — Now persists the fetched mock data onto o._gh (stars, forks, issues, activity) so the filter chip actually has data to evaluate against when a user fetches stats from an org's modal.


S3DFX-CYBER:
@Rishabhworkspace pls follow the pr templates given and mention your program

Issue #1004, opened 2026-05-20
[Enhancement] Make organization cards fully clickable for better UX


malikasingh2509:
### Description

Currently, clicking on an organization display card does not open the organization’s detail page. Users are required to specifically click the "View Details" button at the bottom of the card.

### Problem it Solves

This creates a poor user experience because the card visually appears interactive, but most of the card area is non-clickable. Users naturally expect the entire card to be clickable.

### Proposed Solution

Make the entire organization card clickable so that clicking anywhere on the card navigates the user to the organization detail page.

Possible implementation approaches:

* Wrap the entire card component inside a `Link`
* Add `onClick` navigation to the card container

Additionally:

* Add `cursor-pointer`
* Keep hover states consistent for better UX feedback

### Alternatives Considered

* Keeping only the "View Details" button clickable
* Increasing the button size instead of making the full card interactive

However, making the complete card clickable provides a more intuitive and modern user experience.

### Additional Context

This pattern is commonly used in modern web applications where cards act as interactive navigation elements.

ThakuraAnki:
please assign

S3DFX-CYBER:
@shashanki-hub could u show a mockup if possible so it becomes easier to know how ur going to implement even a code snippet would help , @deepak0x kindly review this issue

shashanki-hub:
> [@shashanki-hub](https://github.com/shashanki-hub) could u show a mockup if possible so it becomes easier to know how ur going to implement even a code snippet would help , [@deepak0x](https://github.com/deepak0x) kindly review this issue

Sure! I’m thinking of making the entire organization card clickable using a wrapper click handler while keeping buttons/actions separately accessible. ill also ensure keyboard accessibility and smooth navigation behavior.

I’ll share a simple mockup/code approach shortly

shashanki-hub:
> [@shashanki-hub](https://github.com/shashanki-hub) could u show a mockup if possible so it becomes easier to know how ur going to implement even a code snippet would help , [@deepak0x](https://github.com/deepak0x) kindly review this issue

Possible implementation:
card.addEventListener("click", () => {
   openOrgDetails(org);
});

ill also:
add cursor: pointer
preserve button-specific actions

deepak0x:
hey @shashanki-hub u can show using some photo or video
this way we can approve this issue easily for you ;)

deepak0x:
[...]

PR #1011, opened 2026-05-20
organization cards fully clickable for better UX #1004

ThakuraAnki:
# ✨ Feature Request: Make Organization Cards Fully Clickable

## 📌 Description

Currently, users can only navigate to an organization’s detail page by clicking the **“View Details”** button located at the bottom of each organization card.

However, the entire card visually appears interactive, leading users to naturally expect that clicking anywhere on the card will open the organization details.

## 🚨 Problem Statement

The current interaction creates an inconsistent user experience because:

* Most of the card area is non-clickable
* Users may repeatedly click on the card itself expecting navigation
* The interface feels less intuitive compared to modern UI standards

This results in unnecessary friction and reduces overall usability.

## ✅ Proposed Solution

Make the **entire organization card clickable** so users can access the organization detail page by clicking anywhere on the card.

### Possible Implementation Approaches

* Wrap the entire card component inside a `Link`
* Add `onClick` navigation to the card container
* Add `cursor-pointer` styling for better interaction feedback
* Preserve or enhance hover animations/states for consistency

## 🎯 Benefits

* Improves overall user experience
* Makes navigation faster and more intuitive
* Aligns with modern web application design patterns
* Enhances accessibility and discoverability
* Creates a smoother and more polished interaction flow

### 1. Keep Only the Button Clickable

This maintains the current behavior but does not solve the usability issue.

### 2. Increase Button Size

While slightly improving accessibility, it still leaves most of the card non-interactive.

Compared to these approaches, making the full card clickable provides the cleanest and most user-friendly experience.

## 📷 Additional Context

Fully clickable cards are a widely adopted pattern in modern applications and dashboards, where cards serve as both informational and navigational UI elements.

This enhancement would make the interface feel more natural, responsive, and professional.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2089,9 +2089,9 @@
       const isBookmarked = bookmarkedSet.has(org.name);
       const isComparing = compareList.includes(org.name);
       const card = document.createElement('article');
-      card.className = 'group bg-white rounded-2xl p-6 border border-zinc-100 transition-all hover:shadow-xl hover:border-primary/20 animate-fade-up';
-      card.dataset.org = org.name;
-      
+   card.className = 'group bg-white rounded-2xl p-6 border border-zinc-100 transition-all hover:shadow-xl hover:border-primary/20 animate-fade-up cursor-pointer';
+card.dataset.org = org.name;
+card.dataset.openOrg = org.name; 
       const githubOwner = githubOwnerFromValue(org.github);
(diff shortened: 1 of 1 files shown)


deepak0x:
hi @ThakuraAnki 
fix the issues reported by @coderabbitai

PR #1118, opened 2026-05-21
fix: null guard in renderCompare() for stale localStorage org names

### LINKED ISSUE #1052: [BUG] renderCompare() crashes with TypeError when compareList contains stale org name from localStorage
## 🐛 Describe the bug
`renderCompare()` throws an uncaught `TypeError: Cannot read properties of undefined (reading 'github')` when `compareList` contains an org name that no longer exists in `ORGS`. The entire compare section breaks silently with no error message shown to the user. This affects all returning users whose compare list was saved in a previous session — if any saved org name doesn't match the current dataset, the feature crashes on every page load with no recovery path.

---

## 🔁 Steps to Reproduce
1. Go to https://findmygsoc.vercel.app
2. Open browser DevTools → Console tab
3. Paste and run this JS to simulate a stale entry:
```js
   globalThis.compareList.push('NonExistentOrg2024');
```
4. Trigger any filter change or reload the page

## ✅ Expected Behavior
The compare section should gracefully skip stale entries and show empty placeholder slots. A message like "One of your saved orgs is no longer available" should be shown so the user understands what happened.

---

## ❌ Actual Behavior
The entire compare section goes blank and crashes silently. No error or message is shown to the user. The feature remains broken for the rest of the session until the user manually clears localStorage — which most users will never think to do.

Root cause in `index.html` → `renderCompare()` around line 2196:
```js
compareList.forEach(name => {
  const org = ORGS.find(o => o.name === name); // returns undefined for stale names
  const githubOwner = githubOwnerFromValue(org.github); // ← TypeError crashes here
```

Fix:
```js
compareList.forEach(name => {
  const org = ORGS.find(o => o.name === name);
  if (!org) return; // ✅ null guard

  const githubOwner = githubOwnerFromValue(org.github);
});

// Also prune stale names on init:
compareList = compareList.filter(name => ORGS.some(o => o.name === name));
```

---

## 🌱 Contributor Checklist

---

## 💻 Environment
- Browser: All browsers (logic bug, not rendering-specific)
- Device: Desktop and Mobile
- OS: All

---

## 📝 Additional Context
This is a silent crash with no user-facing feedback, making it especially hard to diagnose. Any user who added orgs to compare in a previous session and then returns after a data update will hit this bug automatically on page load. The fix is a one-line null guard plus a one-line init-time pruning of stale names.

Rudrasamadhiya:
## 🎯 Program
GSSoC'26

## 📋 Description
Fixes a silent TypeError crash in `renderCompare()` when `compareList` 
in localStorage contains an org name that no longer exists in `ORGS`.

Added a null guard inside the `forEach` loop and a one-line init-time 
pruning of stale names so returning users are never stuck in a broken state.

## 🔗 Related Issue
Closes #1052

## 🔧 Type of Change
- [x] Bug fix (non-breaking fix for an existing issue)

## 🧪 Changes Made
1. Added `if (!org) return;` guard inside `compareList.forEach()`
2. Added `compareList = compareList.filter(name => ORGS.some(o => o.name === name));` 
   on init to prune stale entries before `renderCompare()` runs

**File changed:** `index.html` (~line 2196)

## 🧪 Testing Steps
1. Open https://findmygsoc.vercel.app
2. Open DevTools Console
3. Run: `globalThis.compareList.push('NonExistentOrg2024');`
4. Trigger a filter change
5. ✅ Compare section renders normally, stale entry is silently skipped
6. ✅ No TypeError in console

## ✅ Checklist
- [x] I am a GSSoC'26 participant
- [x] I was assigned before starting work
- [x] PR is linked to the issue with `Closes #1052`
- [x] Changes are minimal and focused
- [x] Commit includes DCO sign-off (`git commit -s`)
- [x] No new dependencies added
- [x] Tested locally


deepak0x (dismissed):
Hi @Rudrasamadhiya, the null guard looks right and it does stop the crash from #1052. One thing still needs a fix though.

The `compareList = compareList.filter(...)` line sits inside the `forEach` loop, so it re-runs on every item instead of once. Your PR description says the pruning should happen at init, before renderCompare runs, but the code puts it inside the render loop instead. The inner filter also reuses `name`, which shadows the loop variable and makes it harder to read. Move the pruning out of the loop.

Separately, the Program Classification check is failing. Please update the PR description so your program is in the format the repo expects.

CodeRabbit and cubic both flagged the filter placement too. Sort that out and I'll take another look.

itsdakshjain (commented):
the null guard handles the issue perfectly.

just wanted to follow up on what deepak and the ai tools noticed earlier—the `compareList.filter` line is stuck inside the `forEach` loop. if you pull it out and drop it just before the loop starts, it'll run a lot better and clean up the code readability.

also, the automated check is hanging because of the program classification tag. if you could just check that box or update the format in your description, the pipeline should go green.

nitinog10 (approved):
Overall, this is a solid defensive fix with good attention to frontend state handling. Nice work catching the edge case.
we can merge after verifying all the tests

S3DFX-CYBER:
Closing due to inactivity

PR #1137, opened 2026-05-22
fix: prevent race condition and TypeError in showMoreIssues()

### LINKED ISSUE #1126: Unhandled null reference crash when Load More button clicked during rapid filter changes
### Description
The `showMoreIssues()` function in `app.js` performs an unsafe DOM query followed by manipulation without null checking. When the issues container DOM is rapidly updated—such as when a user applies filters while a "Load More" operation is pending—the `.issues-grid` child element may be removed or replaced before the function executes. 

This causes `querySelector()` to return null, and the subsequent call to `.insertAdjacentHTML()` throws an unhandled `TypeError`, crashing the issues pagination feature. The failure is silent with no user-facing error message or fallback mechanism.

### Steps to Reproduce
1. Navigate to the Issues page in the application.
2. Click the **"Load Issues"** button to fetch and display good first issues from all organizations.
3. Wait until the issues grid is populated and the "Load More" button becomes visible.
4. Rapidly perform the following actions in sequence:
   * Click the **"Load More"** button.
   * Immediately type a search query in the issue search field (triggering `filterIssues()`).
5. Observe the browser console for a `TypeError`.
6. Notice the issues grid either disappears, becomes partially rendered, or the application enters an unresponsive/undefined state.

### Expected Behavior
* The "Load More" operation should complete successfully, appending the next batch of issues to the grid.
* Alternatively, if a filter is applied during pagination, the operation should gracefully cancel and the filter re-render should take precedence.
* No JavaScript exceptions should be thrown to the console regardless of the order of rapid user interactions.
* The issues grid should always display a consistent set of filtered results.
* The "Load More" button visibility should accurately reflect whether additional paginated results exist.

### Actual Behavior
* Clicking "Load More" while filters are being applied results in an uncaught `TypeError: Cannot read property 'insertAdjacentHTML' of null`.
* The issues grid element is either removed from the DOM before `insertAdjacentHTML()` is called, or the container structure is replaced.
* No user-facing error message or notification is displayed.
* The application silently fails to add more issues and may become partially unresponsive, requiring a page reload or manual intervention to recover.

### Environment
* **Browser:** [e.g., Chrome 125, Firefox 123, Safari 17]
* **OS:** [e.g., Windows 11, macOS 13, Ubuntu 22.04]
* **Application Version:** [e.g., v1.0.0, main branch commit hash]
* **Network Conditions:** [e.g., Normal / 3G / Offline simulation] 

### Proposed Solution
The `showMoreIssues()` function must defensively check that the `.issues-grid` element exists in the DOM before attempting DOM manipulation. Add a null guard clause that safely returns early if the grid element has been removed. 

Additionally, consider implementing a flag or promise-based mechanism to prevent concurrent `showMoreIssues()` and `filterIssues()` operations from executing simultaneously.

**Code Reference:** `app.js`
```javascript
function showMoreIssues() {
  const container = document.getElementById('issuesContainer');
  const next = filteredIssues.slice(shownIssues, shownIssues + ISSUES_PAGE_SIZE);
  shownIssues += next.length;
  container.querySelector('.issues-grid').insertAdjacentHTML('beforeend', next.map(renderIssueCard).join(''));
}

PrincePundir123:
program: gssoc

## 📝 Description
This PR fixes a critical race condition in the `showMoreIssues()` function that causes a silent application crash and corrupts the pagination state when a user rapidly applies filters while a "Load More" operation is pending.

### 🛑 The Problem
When a user clicks "Load More" and immediately applies a filter, `filterIssues()` triggers a DOM re-render that replaces the `.issues-grid` element. If `showMoreIssues()` executes during this DOM update, it attempts to call `.insertAdjacentHTML()` on a `null` reference, throwing an uncaught `TypeError`. 

### ✅ The Solution
1. **Defensive DOM Querying:** Fetched both the `container` and `grid` elements at the beginning of the function using optional chaining.
2. **Guard Clause:** Added an early return (`if (!container || !grid) return;`) to gracefully abort the function if the DOM is in flux.
3. **State Protection:** Moved the `shownIssues` increment strictly *after* the guard clause to ensure the pagination state only updates when the DOM is successfully modified.

## 🔗 Related Issue
Closes #1126 

## 🔄 Type of Change
- [x] 🐛 Bug fix (race condition)
- [x] ✨ New feature
- [x] 📝 Documentation update
- [x] 🎨 Style / UI improvement

## ✅ Checklist
- [x] I have tested these changes manually in the browser.
- [x] My code follows the project's existing style.
- [x] There are no new warnings or errors in the console.
- [x] I have signed off my commits.

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -1194,11 +1194,20 @@ function renderIssueCard(iss){
 }
 
 function showMoreIssues(){
+  // Defensive guard: Verify DOM elements exist before manipulation
   const container=document.getElementById('issuesContainer');
+  const grid=container?.querySelector('.issues-grid');
+  if(!container||!grid){
+    return; // Gracefully abort if DOM is unstable (e.g., during filter re-render)
+  }
+  // Fetch next batch of issues and update pagination state
   const next=filteredIssues.slice(shownIssues,shownIssues+ISSUES_PAGE_SIZE);
   shownIssues+=next.length;
-  container.querySelector('.issues-grid').insertAdjacentHTML('beforeend',next.map(renderIssueCard).join(''));
-  document.getElementById('loadMoreWrap').style.display=shownIssues<filteredIssues.length?'flex':'none';
+  // Safely append new issue cards to the grid
+  grid.insertAdjacentHTML('beforeend',next.map(renderIssueCard).join(''));
+  // Update UI: toggle "Load More" button and pagination count
+  const hasMore=shownIssues<filteredIssues.length;
+  document.getElementById('loadMoreWrap').style.display=hasMore?'flex':'none';
   document.getElementById('issShown').textContent=shownIssues;
 }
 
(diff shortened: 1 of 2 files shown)


PrincePundir123:
@S3DFX-CYBER This is ready for your review. Let me know if I need to change anything.

Issue #1311, opened 2026-05-25
Uncaught QuotaExceededError


annrose2277-glitch:
### Description
Uncaught QuotaExceededError in localStorage analytics tracking crashes the frontend. If a user has strict privacy settings blocking third-party storage, or if their browser's local storage is full, the custom analytics script fails synchronously and halts the execution of the main application logic.

### Steps to Reproduce
Go to the live application.

Open browser DevTools and disable local storage (e.g., block cookies/storage in site settings) or manually fill the local storage quota.

Refresh the page.

See error in the console. The "Trending" section and interactive filtering features fail to initialize.

### Expected Behavior
The application should catch storage-related exceptions gracefully. If localStorage is unavailable, the app should bypass the analytics tracking without breaking the core organization filtering functionality.

### Actual Behavior
A synchronous exception is thrown by localStorage.setItem(), crashing the main JavaScript thread and rendering the page unresponsive to user interactions.

### Environment
OS: Windows 11

Browser: Chrome 124

Node.js: N/A

Platform: Live at Vercel

### Additional Context
Wrap all localStorage.getItem() and localStorage.setItem() calls in a try...catch block within app.js to ensure failures degrade gracefully.

annrose2277-glitch:
/assign

Hi @maintainer,

I would like to work on this bug fix under GSSoC 2026. As I'm preparing my proposals for Google Summer of Code 2026, I've been diving deep into this codebase. Fixing this robustly aligns perfectly with the branching and version control workflows I refined recently during the Hack The Stack competition. 

Please assign this issue to me. Thank you!

PR #1367, opened 2026-05-26
fix: resolve Uncaught QuotaExceededError in localStorage analytics

annrose2277-glitch:
This PR addresses a critical frontend crash caused by unhandled synchronous exceptions in localStorage. Previously, if a user had strict privacy settings blocking third-party storage or if their browser's local storage quota was full, native localStorage.setItem() calls would throw an unhandled QuotaExceededError, halting the main JavaScript thread and breaking core interactive filtering features.
closes issue    #1311 

program: nsoc

Changes Proposed in this PR:

Created a Safe Storage Utility: Added src/js/safe-storage.js containing a robust wrapper object (safeStorage) that wraps native setItem, getItem, and removeItem calls in try...catch blocks.

Implemented Graceful Degradation: The utility now returns null or false on failure and logs a non-blocking warning to the console, allowing the main application logic to continue executing smoothly even when storage is restricted.

Project-wide Refactor: Safely replaced all native localStorage calls across app.js, index.html, and other relevant scripts with the new safeStorage utility.

Updated Script Loading Order: Modified index.html to load <script src="src/js/safe-storage.js"></script> before app.js and other inline scripts to ensure the utility is globally available prior to execution.

Testing:

Verified that blocking local storage via browser DevTools no longer crashes the application.

Confirmed that the filtering and UI rendering degrade gracefully without the analytics engine halting the main thread.


itsdakshjain (commented):
great refactor to stop the localstorage crashes but before we merge, can you just check out the critical warnings from the coderabbit and cubic bots? fixing the missing definition errors for safestorage and the json.parse so corrupted cache doesnot crash the page on startup.

once those are cleared up and the tests pass green, we r good to merge

itsdakshjain:
Add the issue number this solves , and the program under which contributed

itsdakshjain (commented):
Cna you look into the bot comments as they are required to resolved and squash it

PR #1554, opened 2026-05-29
fix: clone headers in fetchWithFallback to prevent shared ghHeaders mutation

### LINKED ISSUE #1336: [Bug]: fetchWithFallback in api/github.js mutates shared ghHeaders object on 401, permanently dropping Authorization for all subsequent API calls in the same invocation
## Bug Description

`api/github.js` contains a `fetchWithFallback` helper that is designed to retry a GitHub API request without the `Authorization` header when a `401` response is received. The implementation mutates the shared `ghHeaders` object by calling `delete options.headers.Authorization`. Because every call within a single edge function invocation passes the same `ghHeaders` reference as `options.headers`, the deletion is permanent for the lifetime of that invocation. All subsequent GitHub API calls in the same request lose the auth token silently and fall back to the unauthenticated rate limit (60 req/hr instead of 5000 req/hr).

### Affected file

`api/github.js`

### Relevant code

```js
// ghHeaders is created once per invocation and shared across all fetch calls
const ghHeaders = {
  Accept: 'application/vnd.github.v3+json',
  'User-Agent': 'gsoc-org-finder',
};
if (token) {
  ghHeaders.Authorization = `Bearer ${token}`;
}

const fetchWithFallback = async (url, options) => {
  let res = await fetch(url, options);
  if (res.status === 401 && options.headers?.Authorization) {
    delete options.headers.Authorization;   // <-- mutates ghHeaders directly
    res = await fetch(url, options);
  }
  return res;
};
```

Every call is made as:
```js
fetchWithFallback(url, { headers: ghHeaders })
```

`options.headers` and `ghHeaders` are the **same object reference**. The `delete` permanently removes `Authorization` from `ghHeaders`.

## Steps to Reproduce

1. Deploy with a `GITHUB_TOKEN` that is expired or intentionally invalid.
2. Send a request to `/api/github?user=someuser`.
3. Internally, the handler enters the `while (page <= 3)` loop:
   - Page 1 receives a `401`. `fetchWithFallback` deletes `ghHeaders.Authorization` and retries unauthenticated.
   - Pages 2 and 3 start with `ghHeaders` that has no `Authorization` at all.
   - All three pages make unauthenticated requests, consuming the 60 req/hr quota.
4. Any deployment that uses the standard stats mode (`Promise.all`) is affected similarly: if the first of the two parallel calls triggers the 401-fallback path first, both retry attempts share the mutated `ghHeaders`.

## Expected Behavior

A 401 fallback should retry the **specific failing request** without the token, leaving all other calls in the invocation unaffected. The shared `ghHeaders` object must not be mutated.

## Actual Behavior

The first `401` response causes `ghHeaders.Authorization` to be deleted for the entire invocation. All subsequent GitHub API calls in that request proceed without authentication, silently halving the effective rate limit from 5000 req/hr to 60 req/hr. Under load this causes premature `403 rate limit exceeded` responses for all users sharing that edge instance.

## Suggested Fix

Clone the headers before mutating instead of modifying `ghHeaders` in place:

```js
const fetchWithFallback = async (url, options) => {
  let res = await fetch(url, options);
  if (res.status === 401 && options.headers?.Authorization) {
    // Create a new headers object without Authorization for the retry only
    const retryOptions = {
      ...options,
      headers: { ...options.headers }
    };
    delete retryOptions.headers.Authorization;
    res = await fetch(url, retryOptions);
  }
  return res;
};
```

This ensures the retry is isolated and `ghHeaders` remains intact for all other calls.

## Impact

- **Rate limit degradation:** A single expired-token request silently drops all subsequent calls in the invocation from 5000 to 60 req/hr.
- **Silent failure:** No log line or error response indicates that authentication was lost mid-invocation.
- **Concurrent requests:** The `Promise.all([repoRes, commitsRes])` pattern in the standard stats mode runs both `fetchWithFallback` calls with the same `ghHeaders` object. Under a 401, whichever call resolves first mutates the shared object, racing with the other in-flight retry.
- **Security posture:** The intent to fall back gracefully on invalid tokens is correct, but the implementation leaves all calls in an unintended unauthenticated state rather than only the one that triggered the 401.

## Environment

- Vercel Edge Runtime (`export const config = { runtime: 'edge' }`)
- `GITHUB_TOKEN` set in Vercel environment variables
- Reproducible when the token is invalid, rotated, or scoped incorrectly

anshul23102:
## 🚀 Program
GSSoC

program: gssoc

## 📝 Description

`fetchWithFallback` in `api/github.js` retried a failing 401 request by calling `delete options.headers.Authorization`. Because every call in the same edge function invocation passes the same `ghHeaders` object as `options.headers`, this deletion was permanent for the entire invocation lifetime. All subsequent GitHub API calls silently lost their auth token and fell back to the 60 req/hr unauthenticated rate limit.

The fix creates shallow copies of both `options` and `options.headers` before the retry, so the deletion is strictly scoped to that one failing request. The shared `ghHeaders` object is never modified.

**Before:**
```js
const fetchWithFallback = async (url, options) => {
  let res = await fetch(url, options);
  if (res.status === 401 && options.headers?.Authorization) {
    delete options.headers.Authorization; // mutates shared ghHeaders
    res = await fetch(url, options);
  }
  return res;
};
```

**After:**
```js
const fetchWithFallback = async (url, options) => {
  let res = await fetch(url, options);
  if (res.status === 401 && options.headers?.Authorization) {
    const retryOptions = {
      ...options,
      headers: { ...options.headers },
    };
    delete retryOptions.headers.Authorization;
    res = await fetch(url, retryOptions);
  }
  return res;
};
```

## 🔗 Related Issue
Closes #1336

## 🔄 Type of Change
- [x] 🐛 Bug fix

## 🧪 How to Test

1. Set `GITHUB_TOKEN` in your environment to an expired or invalid token.
2. Send a request to `/api/github?user=someuser`.
3. Observe that the invocation still completes without all pages falling back to unauthenticated requests.
4. Verify that `ghHeaders.Authorization` is still present after the fallback retry by adding a temporary log line after the `fetchWithFallback` call.

Alternatively:

1. Set `GITHUB_TOKEN` to a valid token.
2. Trigger a request where the first page returns 401 (e.g. by temporarily scoping the token incorrectly).
3. Confirm that only the page that triggered 401 retries unauthenticated; pages 2 and 3 still carry the Authorization header.

## ✅ Checklist
- [x] My PR title follows Conventional Commits format (e.g. `fix: clone headers in fetchWithFallback to prevent shared ghHeaders mutation`)

### DIFF
--- a/api/github.js
+++ b/api/github.js
@@ -53,13 +53,19 @@ export default async function handler(req) {
     ghHeaders.Authorization = `Bearer ${token}`;
   }
 
-  // Helper to fallback to unauthenticated request if token is invalid (401)
+  // Helper to fallback to unauthenticated request if token is invalid (401).
+  // A shallow clone of options.headers is used for the retry so that the
+  // shared ghHeaders object is never mutated and all other calls in this
+  // invocation continue to send the Authorization header normally.
   const fetchWithFallback = async (url, options) => {
     let res = await fetch(url, options);
     if (res.status === 401 && options.headers?.Authorization) {
-      // Remove authorization for this retry and all future requests in this invocation
-      delete options.headers.Authorization;
-      res = await fetch(url, options);
+      const retryOptions = {
+        ...options,
+        headers: { ...options.headers },
+      };
+      delete retryOptions.headers.Authorization;
+      res = await fetch(url, retryOptions);
     }
     return res;
   };


Anushreebasics (approved):
LGTM!

TarunyaProgrammer (approved):
LGTM

S3DFX-CYBER (approved):
Lgtm

PR #1629, opened 2026-05-31
fix: add empty state UI when no orgs match filters

### LINKED ISSUE #1556: ux/feat: Add empty state fallback UI when filter results are zero
### 🎯 Problem Statement
Currently, when a user selects a highly specific or restrictive combination of filters (Tech Stack, Language, and Domain), the grid container becomes completely empty if no organizations match the criteria. A completely blank section lacks user feedback, fails accessibility best practices, and can mislead users into thinking the application has broken or crashed.

### 💡 Proposed Solution
Implement a conditional fallback check within the core rendering logic (where the filtered organization array is mapped to HTML). 

If `filteredOrgs.length === 0`:
1. Instead of leaving the layout blank, render a centered, beautifully styled empty state card.
2. Include a friendly feedback message: *"No organizations match your current filters. Try removing a tag or resetting your search!"*
3. Add a clear, functional "Reset Filters" action link directly inside that card to make recovery effortless.

### 🛠️ Implementation Strategy (Zero-Build Philosophy)
- **Logic:** Add a simple conditional statement in the main rendering JS loop testing if the target array length is zero.
- **UI:** Write clean, native HTML/CSS styles embedded directly into the layout without adding any third-party framework code or npm modules.

thakurakanksha288:
Solution (Zero-Dependency Fix):

src/js/app.js: Upgraded the fallback HTML structure inside renderGrid() to use descriptive, structured class names (.empty-state, .empty-state__title, etc.) and integrated accessibility best practices (role="status", aria-live="polite").

Namespace Protection: Renamed a localized fallback class in renderSelectedLanguages() from empty-state to empty-state-langs to ensure zero CSS rule collision.

src/styles.css: Appended a dedicated, fully responsive styling block for the empty state.

Key Features Added:

Grid Isolation: Used grid-column: 1 / -1 to ensure the empty state card centers perfectly across all screen sizes.

Theme Integration: Leveraged existing CSS variables (var(--ink), var(--blue)) to ensure native compliance with light and dark modes.

Micro-interactions: Embedded a subtle sliding fade-in animation (emptyFadeIn) so the fallback transitions smoothly without jarring the user.

Closes #1556


thakurakanksha288:
hi @S3DFX-CYBER, kindly check once again and let me now.

TarunyaProgrammer:
Hi, @thakurakanksha288. Since this is a UI based PR, consider adding before/after ss in description!

TarunyaProgrammer (dismissed):
Also! I still see two unresolved issues that need to be addressed before this can be merged:

1. **Class name mismatch:** The stylesheet was updated to use `.empty-state-langs`, but `index.html` still renders `<span class="empty-state">No languages selected</span>`. This causes the selected languages placeholder to inherit the new empty-state card styles and can break the layout. Please update the HTML to use `.empty-state-langs` consistently.

2. **Keyboard accessibility:** `.empty-state__reset:focus-visible` currently removes the focus outline (`outline: none`). This hides the keyboard focus indicator and is an accessibility issue. Please provide a visible focus style instead of removing it.

Also, please add before/after screenshots to the PR description so the UI changes can be verified more easily.

thakurakanksha288:
. Fix the Class Name Mismatch
The reviewer noted that your CSS is looking for .empty-state-langs, but your HTML is still using the old .empty-state class.

In index.html
Locate the <span> tag for the empty language state and update the class name.

Before:

HTML
<span class="empty-state">No languages selected</span>
After:

HTML
<span class="empty-state-langs">No languages selected</span>
2. Fix the Keyboard Accessibility (a11y)
Removing the outline on :focus-visible makes it impossible for someone navigating with a keyboard (using the Tab key) to see that they have highlighted the reset button. Instead of outline: none, you should give it a clear, visible focus indicator that matches your design.

In your CSS stylesheet
Locate the .empty-state__reset:focus-visible block and replace outline: none with a visible outline.

Before:

CSS
.empty-state__reset:focus-visible {
    outline: none;
    /* or outline: 0; */
}
After (Recommended approach):

CSS
.empty-state__reset:focus-visible {
    /* Gives a nice, clean 2px solid outline with a small gap around the button */
    outline: 2px solid #0066cc; /* Use your project's primary accent color here */
    outline-offset: 2px;
}

thakurakanksha288:
Hi @maintainers kindly review it once again i have resolved all the conflicts

TarunyaProgrammer (dismissed):
@thakurakanksha288, Thanks for working on this. While reviewing the PR, I found a few blocking issues that need to be addressed before merge:

### 1. `renderGrid()` introduces a JavaScript syntax error

The newly added `renderGrid(orgs)` implementation appears to be incomplete/truncated. The template literal and `.map()` callback are not properly closed, and the code unexpectedly falls into:

```js
renderWatchlist();
updateAIInsights();
}
```

This will result in a parsing error and prevent the application JavaScript from loading.

### 2. Dangling conditional block outside any function

The following block is currently placed between `orgMatchesLanguages()` and `matchesFilters()`:

```js
if (pills.size === 0) {
  container.innerHTML = '<span class="empty-state-langs">No languages selected</span>';
  return;
}
```

Since it is not inside a function scope, it is invalid JavaScript and will break execution. It looks like this logic was intended to be inside `renderSelectedLanguages()`.

### 3. Empty-state class name mismatch

The CSS was updated to use:

```css
.empty-state-langs
```

However, `renderSelectedLanguages()` still renders:

```js
<span class="empty-state">
```

As a result, the "No languages selected" state will not receive the intended styling. Please update the generated markup to use the new class name consistently.

Could you address these issues and push an update?

S3DFX-CYBER:
@thakurakanksha288 kindly fix the tests failing

S3DFX-CYBER (changes requested):
This PR adds an empty-state UI when no orgs match the filters — a centered card with an icon, message, and a "Reset Filters" button, plus a renamed `.empty-state-langs` class to avoid CSS collisions. The design (grid-column: 1 / -1, aria-live, fade-in animation, dark-mode handling) is thoughtful and the linked issue (#1556) is referenced.

However, there are blocking structural problems.

**1. Critical sync violation.** The `renderGrid()` function is rewritten in `src/js/app.js` to emit the empty-state markup, but the inline `<script>` in `index.html` (the production runtime) is not touched in this PR. Per the repo's sync rule, filter/render logic must change in both places — otherwise production users still get the old empty behavior and the two copies diverge. The empty-state feature needs to be mirrored into the inline `renderGrid()` in `index.html`.

**2. Possible structural breakage in app.js.** The patch appears to insert a full `renderGrid()` body (the `g.innerHTML = orgs.map(...)` card markup) inside what was previously the `clearAllBookmarks` function, immediately before `renderWatchlist(); updateAIInsights();`. The diff context shows `}` from clearAllBookmarks followed by the new `renderGrid` and then `renderWatchlist()`/`updateAIInsights()` calls that look orphaned. There's also a fragment `if(pills.size===0){ container.innerHTML='<span class="empty-state-langs">…'` inserted without a closing brace before `function matchesFilters(...)`. This looks like the edits landed in the wrong locations / mismatched braces, which would break the file. Please re-check the hunk boundaries carefully — the empty-state code should replace only the empty branch inside `renderGrid`, not be spliced into unrelated functions.

**3. Stray unrelated additions.** `renderGfiBadge`, `isBookmarked`, and the `['categoryFilter','complexityFilter','sortSelect'].forEach(...)` change-listener block appear to be bundled in alongside the empty-state work. If those aren't part of issue #1556, they should be split into their own PRs to keep this focused.

Please fix the structural splicing, mirror the empty-state into index.html, and split out the unrelated helpers.

PR #1316, opened 2026-05-25
feat: frontend hardening — safe rendering, modal a11y, keyboard nav, SW caching & 23-test coverage

Itzzavdheshh:
## 🚀 Program
NSOC & GSSOC

---

## 📝 Description

Complete frontend hardening and refactor pass for the GSoC Org Finder — preserving the existing vanilla HTML/CSS/JS zero-build architecture while fixing security, accessibility, performance, reliability, and test coverage across all major UI surfaces.

**What was wrong:**
- Modal/keyboard/filter logic duplicated across `index.html` and `app.js`
- Dynamic rendering used raw template strings — no consistent escaping or URL validation
- Inline `onclick` handlers throughout `index.html` — untestable and hard to audit
- Modals lacked `role="dialog"`, focus trapping, Escape-key close, and focus restoration
- GitHub stats modal used simulated/random values instead of the real API-backed flow
- Service worker had no versioned cache, no offline fallback, no request guards
- Zero automated test coverage for filtering, sanitization, recommendations, or modal behavior
- `filteredOrgs` empty on init — `renderOrgs(true)` called before population, cards never rendered

**What was fixed — 6 areas:**

- [x] **Logic consolidation** — all modal, menu, keyboard, filter, bookmark, compare, badge, and load-more behavior centralized in `src/js/app.js` via `addEventListener` and event delegation
- [x] **Safe rendering** — `escapeHtml()`, `safeHTML()`, `sanitizeHrefUrl()`, `validateIdeasUrl()` applied consistently across org cards, issue cards, recommendation cards, modal content, labels, URLs, and logo URLs. Browser scan confirms `inlineHandlers: 0`
- [x] **Modal accessibility** — `role="dialog"`, `aria-modal="true"`, `aria-hidden`, `aria-labelledby` on all modals; Escape closes active modal; focus moves to close button on open; Tab/Shift+Tab trapped inside modal; focus restored to trigger element on close; body scroll locked while open
- [x] **GitHub stats** — modal live stats wired to `fetchModalGH` API-backed flow; `Math.random()` use limited to intentional "Surprise Me" feature only
- [x] **Service worker** — versioned cache name, static asset manifest, install-time caching, old cache cleanup, `clients.claim()`, GET/HTTPS-only guards, network-first for API, stale-while-revalidate for static assets, navigation fallback to `index.html`
- [x] **Tests + CI** — 23 tests across 8 files using Node built-in test runner (`node --test`); CI workflow updated; `npm test` added to `package.json`

---

[...]

PR #1720, opened 2026-06-02
fix(sw): guard cache miss in fetch handler to prevent undefined respondWith

### LINKED ISSUE #1692: Bug: sw.js fetch handler passes undefined to respondWith on cache miss, causing TypeError instead of a graceful offline response
## What happened?

`sw.js` uses a Network First strategy: try the network, and fall back to cache if the network is unavailable. The fallback is implemented as:

```javascript
self.addEventListener('fetch', (event) => {
    event.respondWith(
        fetch(event.request).catch(() => {
            return caches.match(event.request);
        })
    );
});
```

`caches.match(event.request)` returns a Promise that resolves to the cached `Response` if found, or `undefined` if the URL is not cached. When the network fails and the requested URL is not in the cache (for example, any API call or an uncached page), `event.respondWith` receives a Promise that resolves to `undefined`.

Calling `event.respondWith(Promise<undefined>)` causes the browser to abort the request with a `TypeError: Failed to fetch` instead of returning a meaningful offline message. The browser shows a raw network error rather than a controlled degraded-mode response.

## Steps to Reproduce

1. Load the app while online so the install cache is populated.
2. Open DevTools Network tab and set the browser to Offline mode.
3. Navigate to any URL that was NOT pre-cached (e.g. any API endpoint like `api/github?repo=foo`).
4. Observe: the fetch fails with `TypeError: Failed to fetch` rather than any fallback response.

## Expected Behavior

When offline and no cache entry exists for the requested resource, the service worker should return a graceful fallback such as the cached `index.html` (for navigation requests) or a constructed error `Response` (for API requests) so the user sees a clean offline message rather than a browser-level network error.

## Actual Behavior

`event.respondWith` receives `undefined` from the resolved Promise, causing the browser to throw `TypeError: Failed to fetch` for any request that hits a cache miss while offline.

## Root Cause Analysis

`caches.match()` resolves to `undefined` on a miss. The `.catch()` callback does not guard against this case before passing the result to `event.respondWith`.

## Suggested Fix

Add a fallback inside the catch handler:

```javascript
self.addEventListener('fetch', (event) => {
    event.respondWith(
        fetch(event.request).catch(async () => {
            const cached = await caches.match(event.request);
            if (cached) return cached;
            // For navigation requests, serve the cached shell page
            if (event.request.mode === 'navigate') {
                return caches.match('./index.html');
            }
            // Return a minimal offline indicator for other requests
            return new Response(
                JSON.stringify({ error: 'You are offline' }),
                { status: 503, headers: { 'Content-Type': 'application/json' } }
            );
        })
    );
});
```

## Affected File

`sw.js`

## Checklist

- [x] Searched existing issues, not a duplicate (issue #1337 covers stale caches, not cache-miss TypeError)
- [x] Read CONTRIBUTING.md guidelines
- [x] Provided clear reproduction steps
- [x] Described expected vs. actual clearly
- [x] No em dashes or double hyphens
- [x] Repository verified as GSSoC

anshul23102:
## Description

The service worker fetch handler passed `event.respondWith(caches.match(event.request))` without guarding against the cache-miss case. `caches.match()` resolves to `undefined` when the requested URL is not in the cache. Calling `respondWith(Promise<undefined>)` causes the browser to abort the request with a `TypeError` instead of returning a controlled offline response. Offline users saw a raw network error rather than any degraded-mode feedback.

## Root Cause

```javascript
self.addEventListener('fetch', (event) => {
  event.respondWith(
    fetch(event.request).catch(() => {
      return caches.match(event.request); // undefined when not cached
    })
  );
});
```

## Related Issue

Closes #1692

## Type of Change

- [x] Bug fix

## Changes Made

- `sw.js`:
  - Wrapped the catch path in a `.then()` that checks the match result.
  - If a cached response exists, return it as before.
  - For navigation requests with no cache entry, return the cached `index.html` shell so the app remains usable offline.
  - For all other uncached requests, return a `503 Response` so the caller receives a real `Response` object instead of an `undefined` rejection.

## Testing Done

1. Load app online, then switch to offline in DevTools.
2. Navigate to a cached URL: returns cached page as before.
3. Navigate to an uncached route: returns `index.html` shell (app stays usable).
4. Trigger an API call offline: returns `503` response instead of TypeError.

## Checklist

- [x] My code follows the style and formatting of this project
- [x] I have tested my changes locally and they work as expected
- [x] There are no merge conflicts with the base branch
- [x] This PR is linked to the correct issue

### DIFF
--- a/index.html
+++ b/index.html
@@ -1284,10 +1284,44 @@
       <div class="space-y-6">
         <div class="bg-white rounded-2xl p-6 border border-zinc-100">
           <h4 class="font-bold mb-4 flex items-center gap-2"><span class="material-symbols-outlined text-primary text-lg">event</span> Upcoming Deadlines</h4>
-          <div class="space-y-4">
-            <div><p class="text-[10px] font-label uppercase tracking-wider text-zinc-400">MAR 31, 2026</p><p class="text-sm font-bold">Proposal Submission Deadline</p><p class="text-xs text-zinc-500">Ensure your final proposals are submitted by 18:00 UTC.</p></div>
-            <div><p class="text-[10px] font-label uppercase tracking-wider text-zinc-400">APR 21, 2026</p><p class="text-sm font-bold">Review Period Starts</p><p class="text-xs text-zinc-500">Mentors begin ranking of proposals.</p></div>
+          <div id="dynamic-deadlines-container" class="space-y-4">
+            <div class="deadline-node" data-date="2026-03-31">
+              <p class="text-[10px] font-label uppercase tracking-wider text-zinc-400">MAR 31, 2026</p>
+              <p class="text-sm font-bold">Proposal Submission Deadline</p>
+              <p class="text-xs text-zinc-500">Ensure your final proposals are submitted by 18:00 UTC.</p>
+            </div>
+            <div class="deadline-node" data-date="2026-04-21">
+              <p class="text-[10px] font-label uppercase tracking-wider text-zinc-400">APR 21, 2026</p>
+              <p class="text-sm font-bold">Review Period Starts</p>
+              <p class="text-xs text-zinc-500">Mentors begin ranking of proposals.</p>
+            </div>
           </div>
+          <p id="deadlines-empty-fallback" class="text-xs text-zinc-400 italic hidden mt-4">No upcoming deadlines.</p>
+          <script>
+            (function runtimeDeadlineFilter() {
+              const now = new Date();
+              const todayUTC = Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate());
+              const container = document.getElementById('dynamic-deadlines-container');
+              if (!container) return;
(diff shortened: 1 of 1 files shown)


anshul23102:
Fixed the residual risk raised by cubic-dev-ai: `caches.match('index.html')` in the navigation fallback branch is now also chained with a `.then()` guard. If the shell is not cached, a 503 Response is returned instead of passing `undefined` to `respondWith`. Updated in the latest push.

anshul23102:
Hi @S3DFX-CYBER, after rebasing this branch onto current main I found that the SW rewrite in PR #1316 already handles the undefined `respondWith` scenario: the fetch handler now uses proper try/catch with separate network-first (for API calls) and stale-while-revalidate (for static assets) strategies, both of which handle cache misses gracefully rather than passing `undefined` to `respondWith`.

I have reset the branch to current `main` to clear the merge conflicts. This PR can safely be closed since the root cause is addressed in the current `main`. Thank you!

PR #1825, opened 2026-06-10
Fix keyboard focus trap for organization modal

### LINKED ISSUE #1388: [BUG] Escape key does not close organization modal after opening via keyboard navigation
## Describe the Bug
When opening an organization details modal using keyboard navigation, pressing the `Escape` key does not consistently close the modal.
This creates accessibility and keyboard navigation issues, especially for users relying on keyboard-only interaction workflows.

## Steps to Reproduce
1. Navigate through organization cards using keyboard navigation
2. Open an organization modal
3. Press the `Escape` key

## Current Behavior
The modal remains open and requires manual mouse interaction to close.

## Expected Behavior
Pressing `Escape` should close the active modal consistently and restore focus appropriately.

## Proposed Fix
- Add keyboard listener handling for `Escape`
- Ensure modal cleanup executes correctly
- Restore previous focus state after modal close
- Maintain accessibility-friendly interaction flow

## Likely Files Affected
- modal handling logic
- frontend interaction scripts
- keyboard accessibility handling

Namish06:
**Description**
Improved keyboard accessibility for the organization details modal by ensuring focus remains trapped within the modal during keyboard navigation.

**Changes Made**

* Added keyboard focus trapping for `orgModal` using `Tab` and `Shift + Tab` navigation
* Ensured focus cycles only within the organization modal while open
* Added focus handling when opening the modal (`tabindex` + `focus()`)
* Preserved existing `Escape` key behavior for closing the modal
* Improved accessibility for keyboard-only users

**Related Issue**
Fixes #1388

program: gssoc

**Type of Change**
Accessibility improvement

**Checklist**
✅ Tested keyboard navigation manually
✅ Verified `Escape` key still closes modal
✅ Verified focus remains inside modal using `Tab` / `Shift + Tab`
✅ Changes are limited to relevant files
✅ No unrelated changes added

### DIFF
--- a/index.html
+++ b/index.html
@@ -3953,8 +3953,15 @@ let previousFocusedElement = null;
   document.body.style.overflow = '';
 
   if (previousFocusedElement) {
+  if (
+    previousFocusedElement.isConnected &&
+    previousFocusedElement.offsetParent !== null
+  ) {
     previousFocusedElement.focus();
   }
+
+  previousFocusedElement = null;
+}
 }
 
   function openRandomOrg() {
--- a/index.html
+++ b/index.html
@@ -3935,6 +3935,8 @@ if(org.ideas){
      
    const modal = document.getElementById('orgModal');
 
+previousFocusedElement = document.activeElement;
+
 modal.classList.add('open');
 modal.setAttribute('tabindex', '-1');
 modal.focus();
@@ -3943,13 +3945,17 @@ document.body.classList.add('modal-open');
 document.documentElement.style.overflow = 'hidden';
 document.body.style.overflow = 'hidden';
   }
-
+let previousFocusedElement = null;
   function closeOrgModal() {
-    document.getElementById('orgModal').classList.remove('open');
-    document.body.classList.remove('modal-open');
-    document.documentElement.style.overflow = '';
-    document.body.style.overflow = '';
+  document.getElementById('orgModal').classList.remove('open');
+  document.body.classList.remove('modal-open');
+  document.documentElement.style.overflow = '';
+  document.body.style.overflow = '';
+
+  if (previousFocusedElement) {
+    previousFocusedElement.focus();
   }
+}
 
   function openRandomOrg() {
     const orgsToUse = filteredOrgs.length > 0 ? filteredOrgs : ORGS;
--- a/index.html
+++ b/index.html
@@ -3933,13 +3933,15 @@ if(org.ideas){
       draftBtn.onclick = () => ProposalBuilder.open(org.name);
     }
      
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3408,6 +3408,35 @@ <h3 class="font-headline text-lg font-bold text-on-surface mb-1 group-hover:text
...
+  const first = list[0];
+  const last = list[list.length - 1];
+
+  if (e.shiftKey && document.activeElement === first) {
+    e.preventDefault();
+    last.focus();
+  } else if (!e.shiftKey && document.activeElement === last) {
+    e.preventDefault();
+    first.focus();
+  }
```
S3DFX-CYBER, line 3438: This focus-trap listener, the `previousFocusedElement` global, and the rewritten `openOrgModal`/`closeOrgModal` must also be mirrored in `src/js/app.js` — otherwise the production inline script (index.html) and its test mirror (app.js) drift out of sync for modal logic.

S3DFX-CYBER (changes requested):
This PR adds keyboard focus-trapping to the organization details modal so Tab/Shift+Tab cycles focus within the open modal, saves and restores the previously-focused element on open/close, and gives the modal a `tabindex="-1"` + `.focus()` on open. It's a genuine accessibility improvement for keyboard-only users.

**Blocking issues:**

1. **Non-conventional title.** `Fix keyboard focus trap for organization modal` uses a capital `F` and no colon. Conventional Commits requires a lowercase type followed immediately by a colon — e.g. `fix: trap keyboard focus within organization modal`. The body even self-certifies the title follows Conventional Commits, but it does not.

2. **CRITICAL sync rule violation.** All of the changes — the new `keydown` focus-trap listener on `orgModal`, the `previousFocusedElement` global, and the rewritten `openOrgModal`/`closeOrgModal` functions — are made only in the inline script of `index.html`. The test mirror at `src/js/app.js` is not touched, so production and tests will diverge. Any modal-logic change must be mirrored in both files.

**Non-blocking notes:**

- `let previousFocusedElement = null;` is declared between two function bodies rather than at the top of the script/module scope. It works because of hoisting of `var`/function declarations around it, but it's fragile — moving it to a clear top-level declaration would be cleaner.
- The focus-trap listener uses `el.offsetParent !== null` to detect visibility, which returns `null` for elements with `position: fixed`. If the modal ever contains fixed-position controls, they'd be skipped by the focusable-element query. Fine for now, but worth a comment.

Please retitle to a conventional `fix:` form and apply the same modal-logic changes to `src/js/app.js` so the two runtime copies stay in sync.

Issue #911, opened 2026-05-18
[BUG] Dark mode text visibility issue and incorrect Python icon


Dhanashripatil12:
## 🐛 Describe the bug
In dark mode, tech stack card names like React and Next.js are not clearly visible due to low text contrast. Also, the Python tech stack displays an incorrect icon/symbol.

## 🔁 Steps to Reproduce
1.Enable dark mode
2.Navigate to the tech stack/cards section
3.Observe the text visibility and Python icon

---

## ✅ Expected Behavior
Tech stack names should be clearly visible in dark mode, and the correct Python icon should be displayed.
---

## ❌ Actual Behavior
Text becomes difficult/invisible in dark mode, and the Python icon is incorrect.
---

## 📸 Screenshots

---

## 🌱 Contributor Checklist
- [x] I am participating via GSSoC
- [x] I have read the contribution guidelines
- [x] I checked for existing issues before creating this

---

## 💻 Environment
- Browser: (e.g. Chrome, Firefox)
- Device: (e.g. Desktop, Mobile)
- OS: (e.g. Windows, macOS)

---

## 📝 Additional Context
UI issue related to dark mode styling and icon mapping.

<img width="1454" height="656" alt="Image" src="https://github.com/user-attachments/assets/cd8cbfd4-7db3-461b-b5fc-c5a749ab8761" />

<img width="1454" height="656" alt="Image" src="https://github.com/user-attachments/assets/e8c1b46c-3fa5-41c4-96f3-69b7c8096c28" />

HarshadSelokar:
Hi @S3DFX-CYBER, I would like to work on this issue. I’ll reproduce the issue locally, analyze the root cause, and provide a clean fix with proper testing across relevant screen sizes/states.

HarshitaaWantsToCode:
Hi Maintainers, I would love to claim this frontend styling and icon mapping issue; I plan to optimize the text contrast utility classes on the tech stack cards to guarantee proper visibility under dark mode themes, while verifying and updating the underlying icon asset mapping to ensure the correct Python symbol is rendered.

Dhanashripatil12:
Hi Maintainers, I would like to work on this issue under GSSoC'26. Please assign it to me.
/assign gssoc26

hema7392:
Hi @Dhanashripatil12 

I would like to work on this bug under GSSoC'26.

[...]

PR #1835, opened 2026-06-11
fix(ui): improve dark mode text visibility and add tech icons (#911)

aayushprsingh:
This PR addresses issue #911 by:
1. Improving text visibility for tech stack tags in dark mode across the grid, watchlist, and modals.
2. Unifying the Python icon to use a consistent capitalized '🐍 Python' format.
3. Adding icons for React (⚛️) and Next.js (▲) to improve their visibility and recognition.
4. Implementing a 'formatTechTag' helper to centralize technology display logic and ensure consistent styling.

Tests performed:
- Verified tag rendering in dark and light modes.
- Checked contrast ratios for dark mode tags.
- Ran existing test suite ('npm test') to ensure no regressions.

### DIFF
--- a/src/js/recommendation-ui.js
+++ b/src/js/recommendation-ui.js
@@ -123,9 +123,10 @@ function buildCardHtml(rec, currentCompareList, currentBookmarkedSet) {
 
     let matchedSkillsHtml = '';
     if (rec.matchedSkills.length > 0) {
-      const skillsList = rec.matchedSkills.slice(0, 4).map(s =>
-        `<span class="px-2 py-0.5 bg-orange-100 dark:bg-orange-900/30 text-orange-700 dark:text-orange-400 rounded text-[9px] font-bold uppercase tracking-wider">${safeEscapeHtml(s)}</span>`
-      ).join('');
+      const skillsList = rec.matchedSkills.slice(0, 4).map(s => {
+        const displaySkill = s.trim().toLowerCase() === 'python' ? '🐍 python' : s;
+        return `<span class="px-2 py-0.5 bg-orange-100 dark:bg-orange-900/30 text-orange-700 dark:text-orange-400 rounded text-[9px] font-bold uppercase tracking-wider">${safeEscapeHtml(displaySkill)}</span>`;
+      }).join('');
       matchedSkillsHtml = `<div class="mt-2 flex flex-wrap gap-1">${skillsList}</div>`;
     }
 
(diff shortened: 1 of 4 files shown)


src/js/recommendation-ui.js
```diff
@@ -123,9 +123,10 @@ function buildCardHtml(rec, currentCompareList, currentBookmarkedSet) {
 
     let matchedSkillsHtml = '';
     if (rec.matchedSkills.length > 0) {
-      const skillsList = rec.matchedSkills.slice(0, 4).map(s =>
-        `<span class="px-2 py-0.5 bg-orange-100 dark:bg-orange-900/30 text-orange-700 dark:text-orange-400 rounded text-[9px] font-bold uppercase tracking-wider">${safeEscapeHtml(s)}</span>`
-      ).join('');
+      const skillsList = rec.matchedSkills.slice(0, 4).map(s => {
+        const displaySkill = typeof formatTechTag === 'function' ? formatTechTag(s) : (s.trim().toLowerCase() === 'python' ? '🐍 Python' : s);
```
S3DFX-CYBER, line 127: This defensive fallback only covers `python` and silently omits the React/Next.js icons when `formatTechTag` is missing from globalThis. Exposing `formatTechTag` on `globalThis` (as is done for `escapeHtml`) would let you drop the fallback entirely and keep behavior consistent across all call sites.

S3DFX-CYBER (changes requested):
This PR improves dark-mode text contrast across org cards, watchlist, compare table, and issue cards by adding `dark:` Tailwind variants, unifies the Python tag with a 🐍 prefix, adds React ⚛️ and Next.js ▲ icons via a new `formatTechTag` helper, and adds hover styles to mentor chips. It's a substantial, well-structured UI pass — and notably the `formatTechTag` logic is correctly mirrored in both `index.html` (inline) and `src/js/app.js`, so the sync rule is respected there.

**Blocking issue:**

The body says it "addresses issue #911" but does not contain a `Closes #911` / `Fixes #911` / `Resolves #911` keyword. A plain reference does not link the issue for auto-closure; the PR body must include the closing keyword.

**Non-blocking notes:**

- The branch currently has merge conflicts (`mergeable: false`) that will need a rebase against `main` before merge.
- In `src/js/recommendation-ui.js` the `formatTechTag` call is guarded with `typeof formatTechTag === 'function' ? ... : (s.trim().toLowerCase() === 'python' ? '🐍 python' : s)`. That fallback hardcodes only `python` and silently drops React/Next.js icons when `formatTechTag` isn't on `globalThis`. Since the other call sites assume `formatTechTag` is available, consider attaching it to `globalThis` (like `escapeHtml`) so the fallback path is never taken.

Please add `Closes #911` to the PR body and rebase. The code changes themselves are solid.

PR #1842, opened 2026-06-11
fix: keep back-to-top button clear of section controls

### LINKED ISSUE #1828: [BUG]: scroll up arrow appears the entire time when scrolling the page and overlaps with scroll buttons in both mobile and desktop view
## 🐛 Describe the bug
The scroll-up (back-to-top) arrow remains visible while scrolling and overlaps with the existing floating navigation controls located at the bottom-right corner of the page. This causes UI clutter and makes the controls appear crowded on both desktop and mobile views.

---

## ✅ Expected Behavior
- The scroll-up arrow should be positioned so that it does not overlap with other floating navigation controls.
- All floating buttons should remain clearly visible and accessible with proper spacing between them.
- The layout should be consistent across desktop and mobile devices.
- The scroll up arrow should not appear unless the user reaches near the end of the page.

---

## ❌ Actual Behavior
- The scroll-up arrow appears alongside the existing floating navigation controls.
- The buttons overlap or are positioned too close to each other in the bottom-right corner.
- The interface looks cluttered and usability is reduced on both desktop and mobile screens.
- The scroll up arrow appears immediately after scrolling the page a little.
---

## 📸 Screenshots
Image

Image

---

## 🌱 Contributor Checklist

---

saurabhhhcodes:
## Summary
- delay the back-to-top button until the reader is near the end of the page
- move the button into the same floating-control spacing system as the section nav controls
- keep the control stack from colliding on desktop and mobile layouts

## Validation
- `node --check src/js/footer.js`
- `npm test`
- `git diff --check`

Fixes #1828

### DIFF
--- a/src/js/footer.js
+++ b/src/js/footer.js
@@ -26,13 +26,23 @@ document.addEventListener('DOMContentLoaded', () => {
 function initFooter() {
   const backToTopBtn = document.getElementById('back-to-top-btn');
   if (backToTopBtn) {
-    window.addEventListener('scroll', () => {
-      if (window.scrollY > 300) {
+    const updateBackToTopVisibility = () => {
+      const scrollTop = window.scrollY || window.pageYOffset || 0;
+      const viewportHeight = window.innerHeight || document.documentElement.clientHeight || 0;
+      const documentHeight = document.documentElement.scrollHeight || 0;
+      const distanceFromBottom = documentHeight - (scrollTop + viewportHeight);
+      const revealThreshold = Math.max(500, Math.round(viewportHeight * 0.75));
+
+      if (distanceFromBottom <= revealThreshold) {
         backToTopBtn.classList.add('show');
       } else {
         backToTopBtn.classList.remove('show');
       }
-    }, { passive: true });
+    };
+
+    window.addEventListener('scroll', updateBackToTopVisibility, { passive: true });
+    window.addEventListener('resize', updateBackToTopVisibility);
+    updateBackToTopVisibility();
 
     backToTopBtn.addEventListener('click', () => {
       window.scrollTo({ top: 0, behavior: 'smooth' });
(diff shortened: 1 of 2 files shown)


S3DFX-CYBER (changes requested):
The intent — delaying the back-to-top button until the reader is near the bottom, and stacking it above the section-nav FABs using the same spacing variables — is the right fix for the collision. The new visibility logic in `src/js/footer.js` is reasonable: computing `distanceFromBottom` against a `max(500, viewportHeight * 0.75)` threshold and wiring both `scroll` and `resize` (plus an initial call) covers the cases that matter. DCO is verified and `Fixes #1828` is present.

Two things to address:

**Use `globalThis` instead of `window` in `src/js/`.** This is a repo rule for files under `src/js/`. `src/js/footer.js` continues to use `window.addEventListener`, `window.scrollY`, `window.pageYOffset`, `window.innerHeight`, and `window.scrollTo`. Swap these for `globalThis` (e.g. `globalThis.addEventListener('scroll', …)`, `globalThis.scrollY`, `globalThis.scrollTo({ top: 0, behavior: 'smooth' })`) to match the rest of the codebase. The `document.documentElement.*` reads are fine as-is.

**Trailing newline removed from `src/styles.css`.** Same as a few other recent PRs — the diff drops the final newline at the end of `styles.css`. Re-add it; POSIX-correct and avoids noisy future diffs.

**Merge conflicts + DCO label.** The `merge-conflicts` and `dco-missing` labels are both on the PR (even though the head commit shows DCO-verified on the diff — the label may be stale). Either way, the branch conflicts with `main` (`mergeable_state: dirty`), so a rebase is needed before merge. Please also confirm the DCO sign-off is on every commit after the rebase.

Also worth a quick check: the `resize` listener isn't marked `{ passive: true }` like the `scroll` listener. `resize` handlers don't block scrolling the way `scroll`/`touchmove` can, so it's not critical, but for consistency you could leave it as-is or add `{ passive: true }` — minor.

Swap to `globalThis`, restore the trailing newline, rebase, and this should be good to go.

PR #1844, opened 2026-06-11
fix: avoid caching partial GitHub profile data

### LINKED ISSUE #1696: [BUG] Vercel Edge Proxy Caches Poisoned/Partial Data on Timeout
🐛 Describe the bug
In the api/github.js Vercel Edge function, the ?user= API fetches repositories across multiple pagination loops. If the GitHub API request times out (using AbortSignal.timeout(5000)) on page 2 or 3, the script catches the error and breaks the loop, but incorrectly saves the incomplete data to the memory cache via safeCacheSet for a full hour.

🔁 Steps to Reproduce

Trigger the AI recommender for a GitHub user with a very large number of repositories (requiring multiple pages).
Simulate a network delay or timeout on the GitHub API endpoint during page 2 of the fetch loop in api/github.js.
The catch block triggers and breaks the loop, skipping the remaining repositories.
See error: The script continues execution and caches the incomplete result via safeCacheSet(cacheKey, result), meaning the user's AI recommendations will be skewed for the next 60 minutes based on partial data.

✅ Expected Behavior
If a pagination fetch fails due to a timeout or network error, the partial results should either be discarded, or if returned to the user, they should not be saved to the cache. This ensures the next request attempts a fresh, full fetch.

❌ Actual Behavior
Partial, inaccurate statistics are permanently cached for 1 hour, poisoning the data used by the AI recommender.

📸 Screenshots
(Not applicable for backend logic)

🌱 Contributor Checklist
 I am participating via GSSoC
 I have read the contribution guidelines
 I checked for existing issues before creating this

💻 Environment
Browser: Backend (Vercel Edge Function)
Device: Server
OS: Node.js/Edge Runtime

📝 Additional Context
This bug compromises the integrity of the AI recommender system. A simple check to ensure page successfully completed before invoking safeCacheSet will fix this.

saurabhhhcodes:
## Summary
- mark user profile fetches as partial when pagination stops early
- skip hour-long cache writes for incomplete or timeout-truncated repo walks
- keep the response honest by exposing a `partial` flag when the data is incomplete

## Validation
- `node --input-type=module < api/github.js`
- `npm test`
- `git diff --check`

Fixes #1696

### DIFF
--- a/api/github.js
+++ b/api/github.js
@@ -81,6 +81,7 @@ export default async function handler(req) {
     try {
       let page = 1;
       let repos = [];
+      let cacheable = true;
       while (page <= 3) {
         try {
           const res = await fetchWithFallback(`https://api.github.com/users/${user}/repos?per_page=100&sort=updated&page=${page}`, { 
@@ -89,15 +90,21 @@ export default async function handler(req) {
           });
           if (!res.ok) {
             if (page === 1) return new Response(JSON.stringify({ error: `GitHub ${res.status}` }), { status: 502, headers });
+            cacheable = false;
             break;
           }
           const pageRepos = await res.json();
           repos = repos.concat(pageRepos);
           if (pageRepos.length < 100) break;
+          if (page === 3) {
+            cacheable = false;
+            break;
+          }
           page++;
         } catch (e) {
-          // Gracefully break loop on timeout/err for pages 2-3, allowing partial results
-          if (page === 1) throw e; 
+          // Gracefully break loop on timeout/err for pages 2-3, but avoid caching partial data.
+          if (page === 1) throw e;
+          cacheable = false;
           break;
         }
       }
@@ -137,10 +144,11 @@ export default async function handler(req) {
         topics,
         stars: totalStars,
         activity,
+        partial: !cacheable,
         ts: Date.now()
       };
       
-      safeCacheSet(cacheKey, result);
+      if (cacheable) safeCacheSet(cacheKey, result);
       return new Response(JSON.stringify(result), { status: 200, headers });
 
     } catch (err) {


S3DFX-CYBER (commented):
The fix is well-targeted: marking the repo walk as non-cacheable when pagination stops early (non-ok response on pages 2–3, a caught timeout/error, or hitting the page-3 cap) and then gating `safeCacheSet` on `cacheable` is the correct way to prevent partial/truncated profiles from poisoning the edge cache for an hour. Exposing `partial: !cacheable` on the response object is a nice touch — it lets the client be honest with the user about incomplete data rather than silently serving a stale-looking "complete" result. DCO verified, `Fixes #1696` present, conventional `fix:` title, single file (`api/github.js`), no lockfile or dependency changes. This is a clean, focused change.

One correctness nit worth a follow-up rather than blocking:

**Exactly-300-repos cohort marked partial.** The completeness check at the page-3 branch sets `cacheable = false` unconditionally when `page === 3`. But if a user has *exactly* 300 public repos, page 3 returns a full 100 items and the loop exits via the `page === 3` branch — even though the data is actually complete. Those users get marked `partial: true` and never cached, so every request for them re-hits the GitHub API three times. The distinction the code can't make is "page 3 returned exactly 100 (maybe more)" vs "page 3 returned <100 (definitely done)". A cheap fix: after the loop, check whether the last fetched page returned fewer than 100 items — if it did and you broke out via the `< 100` path, the data is complete even if you were on page 3. The current structure makes that a little awkward since the `< 100` break happens before the `page === 3` check, so reordering or tracking a `lastPageComplete` flag would let you keep `cacheable = true` for the exact-300 case.

This is an edge case (very few users have 300+ public repos), so it doesn't block merge in my view — but it's worth a follow-up issue or a small follow-up commit so that cohort stops being uncached.

PR #1846, opened 2026-06-11
fix: tighten timeline hero layout

### LINKED ISSUE #1653: feat: Implement Scrollable Timeline Container with Current Milestone Focus
## Description
The homepage timeline component currently occupies a very large portion of the hero section, which affects both usability and overall page layout.

The timeline panel is significantly taller than necessary and displays many past and future events at once. As a result, the most relevant information—the current active milestone—is not immediately visible and users must scroll through the timeline to locate it.

Additionally, the oversized timeline panel constrains the main content area on the left side of the hero section, making the layout feel unbalanced and leaving large amounts of unused whitespace on the page.

---

## Suggested Improvements

- Set a fixed height for the timeline container.
- Enable internal scrolling for timeline events.
- Auto-scroll to the current active milestone on page load.
- Optimize the layout so both the hero content and timeline section are proportionally sized.
- Reduce unnecessary whitespace created by the oversized timeline component.

---

## Benefits

- Current milestone becomes immediately visible.
- Faster access to the most relevant timeline information.
- Better use of available screen space.
- Improved readability and visual balance.
- More effective presentation of homepage content.
- Cleaner and more user-friendly experience on both desktop and mobile devices.

---

## Current Behaviour

- The timeline panel takes up a large amount of vertical space.
- Users must manually scroll through multiple events to locate the current milestone.
- The timeline begins from older events rather than focusing on the current stage.
- The hero section appears visually unbalanced.
- The main content area is constrained because of the large timeline panel.
- Excess whitespace appears in the layout.

---

## Expected Behaviour

- Reduce the visible height of the timeline component.
- Keep the timeline scrollable within its container.
- Automatically position the timeline so the current/active milestone is visible near the center when the page loads.
- Allow users to scroll upward to view previous milestones and downward to view upcoming milestones.
- Improve the overall balance of the hero section by preventing the timeline from occupying excessive space.
- Allow the main content area to utilize the available space more effectively.

---

## Screenshots

Image
Image

---

## 🌱 Contributor Checklist

saurabhhhcodes:
## Summary
- give the timeline panel a fixed, responsive height so it doesn't swallow the hero
- keep the timeline internally scrollable instead of stretching the page
- center the current milestone on load so the most relevant date is visible immediately

## Validation
- `git diff --check`
- `npm test`

Fixes #1653

### DIFF
--- a/index.html
+++ b/index.html
@@ -1335,7 +1335,7 @@ kbd:hover{
             <h3 class="font-headline font-extrabold text-lg sm:text-xl">2026 Timeline</h3>
             <span class="px-3 py-1 bg-green-100 text-green-700 rounded-full text-[10px] font-bold uppercase tracking-wider">Live</span>
           </div>
-          <div id="timeline-milestones" aria-label="GSoC timeline milestones" class="space-y-4 sm:space-y-5 overflow-y-auto max-h-[50vh] pl-4 no-scrollbar focus:outline-none focus:ring-2 focus:ring-primary/30 rounded-lg"></div>
+          <div id="timeline-milestones" aria-label="GSoC timeline milestones" class="space-y-4 sm:space-y-5 overflow-y-auto pl-4 no-scrollbar focus:outline-none focus:ring-2 focus:ring-primary/30 rounded-lg"></div>
           <div class="mt-5 sm:mt-6 pt-5 sm:pt-6 border-t border-zinc-100">
             <div class="flex items-center justify-between">
               <span class="text-xs text-zinc-500" id="countdown-label">Next milestone</span>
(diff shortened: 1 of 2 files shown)


S3DFX-CYBER (changes requested):
The approach here is sensible — giving `#timeline-milestones` a clamped height via `clamp(18rem, 42vh, 28rem)` so the panel stops stretching the hero, and switching the centering logic to `scrollIntoView({ block: 'center' })` with a `typeof === 'function'` guard is cleaner than the manual offset math it replaces. DCO is verified and `Fixes #1653` is present. No lockfile or dependency changes, and `index.html`/`src/js/app.js` sync isn't implicated since the JS lives inline in `index.html` and the change is self-contained there.

**One real issue: `scroll-behavior: smooth` needs a reduced-motion guard.** `src/styles.css` line 374 sets `scroll-behavior: smooth` on `#timeline-milestones` with no `prefers-reduced-motion` override. For users with vestibular sensitivity or motion-reduction enabled, forced smooth scrolling is an accessibility regression (it's also a WCAG 2.1 SC 2.3.3 concern). Please add:

```css
@media (prefers-reduced-motion: reduce) {
  #timeline-milestones { scroll-behavior: auto; }
}
```

**Trailing-newline removal in `src/styles.css`.** The diff also deletes the final newline at the end of `styles.css` (the `-\n` at the bottom of the hunk). That's a cosmetic regression — POSIX wants a trailing newline, and removing it can cause noisy diffs in tools that expect one. Re-add the trailing newline.

**`merge-conflicts` label.** The branch currently conflicts with `main` (`mergeable_state: dirty`), so a rebase is needed before this can land even after the CSS fix.

Fix the reduced-motion guard, restore the trailing newline, and rebase — then this should be good.

PR #1870, opened 2026-06-13
fix: handle cache miss in sw.js fetch handler with proper offline fallback

### LINKED ISSUE #1692: Bug: sw.js fetch handler passes undefined to respondWith on cache miss, causing TypeError instead of a graceful offline response

saurabhhhcodes:
When offline and a URL is not in cache, sw.js passed undefined to respondWith causing TypeError. Now returns a graceful offline fallback — cached index.html for navigation requests, JSON error for others.

Closes #1692

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/1870?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


S3DFX-CYBER (changes requested):
Thanks for the offline fallback work — the intent here is solid, but there's a regression I can't merge past.

**Duplicate `install` handler (blocker).** `sw.js` now registers two separate `globalThis.addEventListener('install', …)` listeners. The original one at the top of the file already runs on install, and this PR adds a second one starting at line 46 that re-does the `caches.open(CACHE_NAME)` → `addAll(CRITICAL_ASSETS)` → `Promise.all(OPTIONAL_ASSETS.map(fetch))` → `skipWaiting()` flow. That means every install runs precache + optional fetch logic twice — doubled network and cache work on every SW lifecycle event, and two `skipWaiting()` calls fighting each other. If you genuinely need the optional-asset precaching, fold it into the single existing install handler rather than adding a second listener.

**Merge conflicts + DCO.** The branch has merge conflicts against `main` and is missing the DCO sign-off (the `dco-missing` label is on). You'll need to rebase and re-sign with `git commit -s --amend` before this can land.

**Title / issue linkage.** Title is conventional (`fix:`) which is good. `Closes #1692` is present in the body, so the issue linkage is fine.

A couple of smaller things on the fetch handler itself: the offline JSON fallbacks now correctly carry `status: 503` (good), but the inline HTML fallback string on the navigate branch is a long single-line blob — consider lifting it to a module-level constant or a tiny `OFFLINE_HTML` so the handler stays readable. Also the redundant `response.status === 200 || response.ok` check on the optional-asset fetch — `response.ok` already covers 200, so the `=== 200` half is dead.

Resolve the duplicate install handler, rebase cleanly, sign off, and we can take another look.

PR #1885, opened 2026-06-16
fix: handle QuotaExceededError in bookmark localStorage writes

### LINKED ISSUE #1697: [BUG]  Unhandled QuotaExceededError in Bookmark Synchronization
🐛 Describe the bug
The local storage bookmarking logic does not handle quota limitations. In index.html, the syncBookmark function blindly writes the bookmarkedSet to localStorage via localStorage.setItem('bookmarks', ...) without a try/catch block. If the browser blocks storage or the domain quota is full, this throws an uncaught error, crashing the script and silently failing to save the user's bookmark.

🔁 Steps to Reproduce

Go to the web app in strict Incognito mode or intentionally fill the domain's local storage quota (e.g., via the console).
Click on the 'Star/Bookmark' icon on any organization card.
Open the browser console.
See error: Uncaught SecurityError or QuotaExceededError, and the UI state fails to persist.

✅ Expected Behavior
The bookmark toggle should be wrapped in a try/catch block. If a quota error occurs, the application should gracefully catch it and alert the user that their browser settings or storage limits prevent saving bookmarks.

❌ Actual Behavior
The script crashes entirely and the user is unaware that their bookmark was not actually saved until they refresh the page.

📸 Screenshots
n/a

🌱 Contributor Checklist
 I am participating via GSSoC
 I have read the contribution guidelines
 I checked for existing issues before creating this

💻 Environment
Browser: Safari (Strict Privacy Mode), Chrome
Device: Desktop, Mobile
OS: macOS, Windows, iOS

📝 Additional Context
The application already has a saveCache function that attempts to handle quota limits for the API responses; similar resilience needs to be applied to user bookmarks.

saurabhhhcodes:
## Description
Wraps `localStorage.setItem` in `syncBookmark` and `clearAllBookmarks` with try/catch to prevent uncaught `QuotaExceededError` / `SecurityError` when storage is unavailable or full.

**Changes**
- `syncBookmark` — catch storage write failure, alert user, return early
- `clearAllBookmarks` — catch storage write failure, alert user, return early

**Fixes** #1697

## GSSoC
- Closes #1697

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/1885?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -433,8 +433,6 @@ kbd:hover{
     .proposal-preview blockquote { border-left: 4px solid #ddd; padding-left: 1rem; color: #666; margin-bottom: 1rem; }
     .proposal-preview a { color: #f97316; text-decoration: underline; }
     .proposal-preview-error { color: #ba1a1a; font-size: 0.875rem; }
-    .sr-only { position: absolute; width: 1px; height: 1px; padding: 0; margin: -1px; overflow: hidden; clip: rect(0, 0, 0, 0); white-space: nowrap; border: 0; }
-
     /* Custom Scrollbar for Modal */
     .modal::-webkit-scrollbar { width: 6px; }
     .modal::-webkit-scrollbar-track { background: transparent; }
--- a/index.html
+++ b/index.html
@@ -3651,7 +3651,12 @@ btn.__orgListenerAttached = true;
     if (!orgName) return;
     if (shouldAdd) bookmarkedSet.add(orgName);
     else           bookmarkedSet.delete(orgName);
-    localStorage.setItem('bookmarks', JSON.stringify([...bookmarkedSet]));
+    try {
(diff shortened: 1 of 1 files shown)


itsdakshjain (commented):
before we can merge

revert the css please put .sr-only back, it looks like it was deleted by accident.

fix dco sign-off 

remove alert using alert() fails our sonarcloud check. please use a non-blocking notification or console error instead.

state bug  you are changing the bookmark set before the try/catch. if storage fails, the ui and localstorage will get out of sync.

PR #1931, opened 2026-06-21
fix: add hover effects to mentor section buttons (#1318)

saurabhhhcodes:
Fixes #1318

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/1931?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1930, opened 2026-06-21
fix: add meaningful alt text to organization logo images (#1200)

saurabhhhcodes:
Fixes #1200

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/1930?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1929, opened 2026-06-21
fix: add target="_blank" and rel to privacy policy footer link (#1576)

saurabhhhcodes:
Fixes #1576

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/1929?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #1932, opened 2026-06-21
fix: remove duplicate back-to-top button (#1894)

### LINKED ISSUE #1894: [BUG] : Remove duplicate two back to top buttons
## Issue Description
The UI currently displays multiple "Back to Top" floating buttons on the page, resulting in visual clutter and an inconsistent user experience. As shown in the screenshot, two similar floating action buttons are rendered simultaneously and appear misaligned with the section indicator component.

## Current Behaviour

- Two "Back to Top" buttons are visible at the same time.
- Floating buttons are not properly aligned with each other.
- The duplicated controls create confusion about their functionality.
- The right-side floating action area appears overcrowded.

## Expected Behaviour

- Only one Back to Top button should be displayed.
- The button should be consistently positioned across all screen sizes.
- Proper spacing and alignment should be maintained between the section indicator and floating action button.
- The floating action area should remain clean and intuitive.

## Steps to Reproduce

- Navigate to any page with the floating navigation controls enabled.
- Scroll down the page.
- Observe the right-side floating action area.
- Notice the presence of multiple Back to Top buttons and their misalignment.

## Proposed Solution

- Identify and remove the redundant Back to Top component.
- Ensure a single source of truth for scroll-to-top functionality.
- Refactor positioning styles (fixed, bottom, right, z-index) to prevent overlapping elements.
- Verify responsive behaviour on desktop, tablet, and mobile devices.
- Add UI consistency checks to prevent duplicate floating actions from rendering.

Image

saurabhhhcodes:
Fixes #1894

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/1932?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -433,8 +433,6 @@ kbd:hover{
     .proposal-preview blockquote { border-left: 4px solid #ddd; padding-left: 1rem; color: #666; margin-bottom: 1rem; }
     .proposal-preview a { color: #f97316; text-decoration: underline; }
     .proposal-preview-error { color: #ba1a1a; font-size: 0.875rem; }
-    .sr-only { position: absolute; width: 1px; height: 1px; padding: 0; margin: -1px; overflow: hidden; clip: rect(0, 0, 0, 0); white-space: nowrap; border: 0; }
-
     /* Custom Scrollbar for Modal */
     .modal::-webkit-scrollbar { width: 6px; }
     .modal::-webkit-scrollbar-track { background: transparent; }
--- a/index.html
+++ b/index.html
@@ -3651,7 +3651,12 @@ btn.__orgListenerAttached = true;
     if (!orgName) return;
     if (shouldAdd) bookmarkedSet.add(orgName);
     else           bookmarkedSet.delete(orgName);
-    localStorage.setItem('bookmarks', JSON.stringify([...bookmarkedSet]));
+    try {
+      localStorage.setItem('bookmarks', JSON.stringify([...bookmarkedSet]));
+    } catch (e) {
+      alert('Could not save bookmark due to storage limits. Please free up some space in your browser settings.');
+      return;
+    }
     refreshOrgGridAfterBookmarkChange();
     renderWatchlist();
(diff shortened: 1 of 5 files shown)


S3DFX-CYBER (changes requested):
This removes the duplicate back-to-top button from `src/components/footer.html` and its initialization logic from `src/js/footer.js`, linked to issue #1894. The footer had its own back-to-top button (`#back-to-top-btn`) in addition to the one in `index.html`, causing a duplicate. Removing the footer copy is the right fix.

However, this PR also includes several unrelated changes that appear across #1930, #1929, #1931, and #1934: the bookmark `try/catch` in index.html, the privacy link `target="_blank"`, the twitter hover color change, the `.sr-only` class removal, and the `isContentEditable` keyboard handler guard. These are all separate fixes that should be in their own PRs. Strip them out and keep only the back-to-top button removal.

The back-to-top CSS in `src/styles.css` (`#back-to-top-btn`, `#back-to-top-btn.show`, `#back-to-top-btn:hover`, `#back-to-top-btn:focus-visible`, and the privacy-page overrides) is correctly removed. But verify that the back-to-top button still exists in `index.html` — the footer's copy was the duplicate, and the original should still be present in the main page. If `index.html` references `#back-to-top-btn`, the CSS for it should still exist somewhere. Check whether the CSS was defined in `src/styles.css` (which you're removing it from) or in `index.html`'s inline `<style>` block.

The `src/js/footer.js` change removes the scroll listener and click handler for the back-to-top button. Make sure the button in `index.html` has its own event listeners — otherwise the button will be visible but non-functional.

PR #1934, opened 2026-06-21
fix: align privacy.html navbar and footer styling with homepage theme (#1656)

### LINKED ISSUE #1656: [BUG] privacy.html navbar and footer styling doesn't match homepage theme
## Description
The `privacy.html` page has a different navbar and footer style compared to `index.html`.

## What's Wrong
- Navbar logo size and style is different from the homepage
- Footer styling doesn't match the homepage design

## Expected Behavior
`privacy.html` should have the same navbar and footer styling as `index.html`

## Screenshots

Image
Image

saurabhhhcodes:
Fixes #1656

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/1934?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -433,8 +433,6 @@ kbd:hover{
     .proposal-preview blockquote { border-left: 4px solid #ddd; padding-left: 1rem; color: #666; margin-bottom: 1rem; }
     .proposal-preview a { color: #f97316; text-decoration: underline; }
     .proposal-preview-error { color: #ba1a1a; font-size: 0.875rem; }
-    .sr-only { position: absolute; width: 1px; height: 1px; padding: 0; margin: -1px; overflow: hidden; clip: rect(0, 0, 0, 0); white-space: nowrap; border: 0; }
-
     /* Custom Scrollbar for Modal */
     .modal::-webkit-scrollbar { width: 6px; }
     .modal::-webkit-scrollbar-track { background: transparent; }
--- a/index.html
+++ b/index.html
@@ -5036,7 +5036,8 @@ if (
     active &&
     (
         ['INPUT', 'TEXTAREA', 'SELECT'].includes(active.tagName) ||
-        active.id === 'searchInput'
+        active.id === 'searchInput' ||
+        active.isContentEditable
     )
 ) {
         return;
--- a/index.html
+++ b/index.html
@@ -3651,7 +3651,12 @@ btn.__orgListenerAttached = true;
     if (!orgName) return;
     if (shouldAdd) bookmarkedSet.add(orgName);
     else           bookmarkedSet.delete(orgName);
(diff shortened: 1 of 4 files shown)


S3DFX-CYBER (changes requested):
This aligns the privacy.html navbar and footer styling with the homepage theme — adds a theme toggle button to privacy.html, glass effect on the header, desktop nav link underline animations, dark mode header overrides, and removes the privacy-page-specific footer CSS overrides from `src/styles.css` (replacing them with a comment saying the privacy page now uses the same footer styling as the homepage). Linked issue #1935 is present.

The theme toggle implementation in privacy.html is self-contained — a `toggleTheme()` function with localStorage persistence and `aria-pressed`/`aria-label` updates. This duplicates the theme toggle logic that exists in `index.html` and `src/js/app.js`. If the theme toggle logic ever changes in the main app, this copy in privacy.html will need to be updated too. Consider extracting the theme toggle into a shared `src/js/theme.js` module that both pages import, so there's a single source of truth. That's an architecture improvement, not a blocker for this PR.

The `.sr-only` class is removed from `index.html`'s inline styles. Verify that `.sr-only` is defined elsewhere (e.g., in `src/styles.css` or via Tailwind's `sr-only` utility) — if it was only defined in the inline `<style>` block, removing it will break screen-reader-only elements that use the class.

The `src/styles.css` diff removes the privacy-page-specific footer overrides (`.privacy-page .premium-footer`, `.privacy-page .footer-stat-card`, `.privacy-page .social-link`). This means the privacy page will now use the same footer styling as the homepage. Verify this looks correct — the privacy page has a different background color scheme than the homepage, and the footer may need the overrides to look right.

The `index.html` changes also include the bookmark `try/catch` error handling and the `isContentEditable` check in the keyboard handler — these are the same changes that appear in #1930, #1929, #1932, and #1931. These are unrelated to the privacy.html navbar alignment. Strip them out so this PR is focused on the privacy page changes only.

The `src/js/app.js` change (adding `isContentEditable` to the keyboard handler guard) and the `src/components/footer.html` change (adding `target="_blank"` to the privacy policy link) are also duplicated across multiple PRs. Pick one PR for each of these changes.

The twitter hover color change in `src/styles.css` (from `#1DA1F2` to `#000000`) is also duplicated across #1930, #1929, #1932, and #1931. It belongs in one PR only.

Issue #1945, opened 2026-06-23
feat: System Theme Preference Support


desireddymohithreddy0925:
## ✨ Feature Description
Add support to automatically detect and respect the user's operating system theme preference (dark/light mode).

## 🤔 Problem It Solves
Currently, users must manually click the theme toggle button to switch to dark mode, and this preference is saved in `localStorage`. Users whose operating systems automatically switch themes based on the time of day (or manual OS toggles) have a disconnected experience where the website does not adapt automatically, forcing them to manually toggle it.

## 💡 Proposed Solution
1. Use `window.matchMedia('(prefers-color-scheme: dark)')` to check the system's preferred theme on initial load, using it as the default if `localStorage` has no saved preference.
2. Add an event listener to `matchMedia` to dynamically switch the theme if the system preference changes during the session.

## 🔄 Alternatives Considered
Keeping it purely manual (status quo), which is simple but lacks the seamless modern experience users expect.

## 📸 Mockups / Examples
N/A - This is a background logic update that affects the existing UI toggle.

## ✅ Acceptance Criteria
- [x] If no localStorage preference is set, the site loads in the OS's preferred color scheme.
- [x] If the OS theme changes while the site is open, the site updates its theme automatically.
- [x] If the user manually clicks the theme toggle, it overrides the system preference and saves to localStorage.

## 📋 Additional Context
This logic should be added to the early-executing theme script in `index.html` or the initialization phase in `app.js` to prevent FOUC (Flash of Unstyled Content).

## 🌱 Contributor Checklist
- [x] I am participating via GSSoC
- [x] I have read the contribution guidelines
- [x] I checked for existing issues before creating this

PR #1953, opened 2026-06-24
feat: System Theme Preference Support #1945

desireddymohithreddy0925:
### ✨ Feature Description
Added support to automatically detect and respect the user's operating system theme preference (dark/light mode).

### 🤔 Problem It Solves
Currently, users must manually click the theme toggle button to switch to dark mode, and this preference is saved in localStorage. Users whose operating systems automatically switch themes based on the time of day (or manual OS toggles) have a disconnected experience where the website does not adapt automatically, forcing them to manually toggle it.

### 💡 Proposed Solution
- Used `window.matchMedia('(prefers-color-scheme: dark)')` to check the system's preferred theme on initial load in `index.html` and `app.js`, using it as the default if `localStorage` has no saved preference.
- Added an event listener to `matchMedia` to dynamically switch the theme if the system preference changes during the session.
- Prevented FOUC (Flash of Unstyled Content) by adding the fallback check early in the scripts.

### ✅ Acceptance Criteria Verified
- [x] If no localStorage preference is set, the site loads in the OS's preferred color scheme.
- [x] If the OS theme changes while the site is open, the site updates its theme automatically.
- [x] If the user manually clicks the theme toggle, it overrides the system preference and saves to localStorage.

### 🌱 Contributor Checklist

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/1953?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

program: gssoc

### DIFF
--- a/.github/workflows/program-classification-validator.yml
+++ b/.github/workflows/program-classification-validator.yml
@@ -86,35 +86,30 @@ jobs:
               // ------------------------------------
               // Metadata
               // ------------------------------------
+              
+              const normalizedBody = body.replace(/\r\n/g, '\n');
+              const has = (re) => re.test(normalizedBody);
 
               if (
-                body.includes('program: gssoc') ||
-                body.includes('- [x] gssoc') ||
-                body.includes('program\ngssoc') ||
-                body.includes('program\r\ngssoc') ||
-                body.includes('- [x] i am contributing under gssoc') ||
-                body.includes('- [x] i am participating via gssoc')
+                has(/\bprogram\s*:\s*gssoc\b/i) ||
+                has(/^\s*[-*]\s*\[x\]\s*(gssoc|i am contributing under gssoc|i am participating via gssoc)\b/im) ||
+                has(/\bprogram\n\s*gssoc\b/i)
               ) {
                 detected.add('gssoc');
               }
 
               if (
-                body.includes('program: nsoc') ||
-                body.includes('- [x] nsoc') ||
-                body.includes('program\nnsoc') ||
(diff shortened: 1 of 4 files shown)


S3DFX-CYBER (changes requested):
This adds system theme preference support — when no saved theme exists in localStorage, the app checks `prefers-color-scheme: dark` and follows the OS preference. It also listens for `matchMedia` changes to update in real-time. The implementation is in both `index.html` and `src/js/app.js`, correctly in sync. The linked issue #1945 is present.

The PR also includes the same `program-classification-validator.yml` workflow regex changes that appear in #1960 and #1961. Those are unrelated to theme support — split them into their own PR.

The `index.html` version uses `window.matchMedia` while `src/js/app.js` uses `window.matchMedia` as well. The repo convention is `globalThis` over `window` in `src/js/` files. Switch to `globalThis.matchMedia` in app.js.

One behavioral note: the `matchMedia` change listener checks `if (!localStorage.getItem('theme'))` before toggling. This means if the user has explicitly set a theme preference, the OS preference change won't override it — which is correct. Good.

Otherwise this is a clean, focused feature (minus the workflow changes). Strip the workflow changes and switch to `globalThis` and it's good.

PR #1965, opened 2026-06-27
feat: derive org years dynamically from firstYear

### LINKED ISSUE #1951: feat: derive org `years` dynamically from firstYear instead of hardcoded value
## ✨ Feature Description

The `years` field on each org entry in `src/js/org.js` is a **hardcoded integer** that must be manually bumped every GSoC cycle. It should be **derived dynamically** from `firstYear` and the current (or latest-known) GSoC year, so the dataset never silently goes stale.

## 🤔 Problem It Solves

Today every org carries a hand-maintained `years` value, e.g.:

```js
{ name: "Apache Software Foundation", firstYear: 2016, years: 11, ... }
{ name: "Boa", firstYear: 2026, years: 1, ... }
{ name: "Git", firstYear: 2016, years: 11, ... }
```

This value is consumed by the recommendation engine in `src/js/recommender.js`:

```js
const years = org.years || 0;
if (years >= 10) { matchReasons.push(`GSoC Veteran (${years} years)`); }
if (years >= 5)  { ... }
```

Because `years` is static, it drifts the moment the calendar rolls over to a new cycle:

- An org with `firstYear: 2016, years: 11` was accurate for **2026** (2016→2026 = 11 cycles).
- In **2027** that same org participates in 12 cycles, but the badge still says *"GSoC Veteran (11 years)"* until every entry is edited by hand.
- Any org added late or skipped in a manual sweep keeps a wrong number indefinitely, so the "GSoC Veteran" / veteran-tier logic in `recommender.js` (lines 131–136) renders incorrect match reasons.

Anyone relying on the recommendation badge — prospective contributors comparing org tenure — sees a stale number with no signal that it is out of date.

## 💡 Proposed Solution

Derive `years` at read time from `firstYear` instead of storing it:

```js
// helper, e.g. in org.js or a small utils module
const LATEST_GSOC_YEAR = 2026; // single source of truth, bumped once per cycle
const participationYears = (org) => (org.firstYear ? LATEST_GSOC_YEAR - org.firstYear + 1 : 0);
```

Then replace `org.years` references (notably `recommender.js:131`) with the computed value, and drop the redundant `years` key from the dataset (or keep it as a deprecated fallback during migration).

A single `LATEST_GSOC_YEAR` constant replaces ~100+ per-org integers, so the annual update becomes a one-line change.

## 🔄 Alternatives Considered

- **Keep `years`, add a CI lint** that fails when `years !== currentYear - firstYear + 1`. This catches drift but still requires editing every entry each year — more churn than deriving the value.
- **Fetch the current year from `new Date().getFullYear()`** at runtime. Rejected because GSoC is a fixed annual program — mid-cycle (Jan–Apr) the "latest" year should stay pinned to the last completed cycle, not roll over on Jan 1. A pinned constant matches the program's real cadence.

## ✅ Acceptance Criteria

- [ ] `years` is computed from `firstYear` + a single pinned "latest GSoC year" constant; the static `years` field is removed (or deprecated with a migration note).
- [ ] `recommender.js` "GSoC Veteran" badge logic uses the computed value.
- [ ] Existing tests in `tests/recommendation.test.js` still pass, with assertions updated to the computed values.
- [ ] No org entry renders a `years` value that is inconsistent with `firstYear`.
- [ ] A brief note added to the README or a code comment on how to bump the constant each cycle.

## 📋 Additional Context

- Consumption site: `src/js/recommender.js` lines 131–136.
- Data source: `src/js/org.js` (the `ORGS` array); related generated stats live under `data/` (`data/org-stats.json`, `data/last-updated.json`).
- Scope is intentionally small: one helper, one constant, update the recommender call site, and adjust tests — no UI changes.

## 🌱 Contributor Checklist

kashyapanshu3108-ops:
## Related Issue

Closes #1951

## Program

GSSoC 2026

## Summary

This PR removes the need to maintain hardcoded `years` values for every organization by deriving the participation years from `firstYear`.

### Changes made

- Added a single `LATEST_GSOC_YEAR` constant as the source of truth.
- Added a reusable `participationYears()` helper function.
- Removed hardcoded `years` values from `src/js/org.js`.
- Updated the recommendation logic to use the computed participation years for GSoC veteran badges.

## Why this change?

Previously, the `years` field had to be updated manually every GSoC cycle, which could easily become outdated.

By calculating participation years dynamically from `firstYear`, future maintenance becomes much simpler while keeping the displayed experience consistent.

## Testing

- Verified that organization cards render correctly.
- Verified that search and filtering continue to work.
- Verified that the recommendation feature uses dynamically calculated participation years.
Screenshot 2026-06-27 115342
Screenshot 2026-06-27 115352

- [x] I have tested my changes locally.
- [x] I have followed the contribution guidelines.
- [x] This PR is linked to an existing issue.

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/1965?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


kashyapanshu3108-ops:
The proposed solution derives participation years from firstYear, but the dataset contains organizations that skipped some GSoC cycles. Would you prefer keeping an explicit participation count for those cases, or maintaining a participation history instead?

S3DFX-CYBER:
@kashyapanshu3108-ops tests are failing

PR #1990, opened 2026-07-01
fix(perf): debounce search input to prevent excessive filter re-renders

### LINKED ISSUE #1970: [PERFORMANCE] Search input triggers filter on every keystroke without debouncing

anshul23102:
## Description

The organization search input and mentor search input triggered a full filter/render operation on every single keystroke without debouncing, causing 15+ synchronous filter operations when typing a typical search query, resulting in UI sluggishness and unnecessary DOM thrashing.

## Root Cause

The `input` event listeners for `searchInput` and `mentorSearchInput` called `applyFilters` / `renderMentorFinder` directly with no debounce wrapper.

## Fix

- Added a small `debounce(fn, delay)` utility function.
- Wrapped the `searchInput` listener (and the synced `hero-search` input) with a 300ms debounced call to `applyFilters`.
- Wrapped the `mentorSearchInput` listener with a 300ms debounced call to `renderMentorFinder`.

## Testing

- Verified with `node --check` that the modified file has valid syntax.
- Manually confirmed typing in the search box no longer triggers a filter operation on every keystroke; filtering now runs only after a brief pause in typing.
- No change to filter behavior/output, only to timing of invocation.

Fixes #1970

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/1990?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -49,6 +49,22 @@ const CATEGORY_META = {
   other: { className: 'bg-zinc-100 text-zinc-600', label: 'Other' },
 };
 
+// ══════════════════════════════════════════════
+// UTILITY FUNCTIONS
+// ══════════════════════════════════════════════
+/**
+ * Returns a debounced version of fn that delays invoking it until after
+ * `delay` ms have elapsed since the last call. Prevents excessive filter
+ * re-renders on every keystroke in search inputs.
+ */
+function debounce(fn, delay) {
+  let timer;
+  return function (...args) {
+    clearTimeout(timer);
+    timer = setTimeout(() => fn.apply(this, args), delay);
+  };
+}
+
 const LANGUAGE_ALIASES = {
   'python': ['python'],
   'javascript': ['javascript', 'js'],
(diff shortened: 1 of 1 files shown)


anshul23102:
Hi @S3DFX-CYBER, following up on this PR - it addresses the search debounce issue (#1970) and has been open for a while awaiting review. Could you please take a look when you get a chance, and apply the appropriate GSoC labels if it looks good? Thanks for your time.

S3DFX-CYBER (changes requested):
This adds a debounce utility to `src/js/app.js` and wires it to the search input and mentor search input with a 300ms delay. The implementation is correct — `clearTimeout` before each call, `fn.apply(this, args)` preserves context. But this is a duplicate of #1984, which is the same feature by the same author.

Close this one and keep #1984 — it has a linked issue and uses 250ms which is a better debounce delay for search. Or close #1984 and keep this one, but pick one. Having both open just creates merge conflicts.

There's no linked issue in the PR body. Add a `Closes #N` if you keep this one.

The debounce function only lives in `src/js/app.js`. The `index.html` inline script also wires up search input listeners — if those aren't also debounced, the production runtime will still fire `applyFilters` on every keystroke. Verify the inline script's search listener is updated too, per the index.html/app.js sync rule.

PR #1996, opened 2026-07-05
feat: add dedicated contributing webpage

MUKUL-PRASAD-SIGH:
## Summary
- add a dedicated contributing page for the site
- route the footer contribution link to the new page
- add a regression test for the new page

## Testing
- npm test

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/1996?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/components/footer.html
+++ b/src/components/footer.html
@@ -37,7 +37,7 @@
           <span class="material-symbols-outlined text-base">menu_book</span> Resources
         </a>
         <a class="footer-link text-sm text-zinc-600 hover:text-primary dark:text-zinc-300 dark:hover:text-primary transition-all flex items-center gap-1.5"
-          href="https://github.com/S3DFX-CYBER/GSoC-Org-Finder-/blob/main/CONTRIBUTING.md">
+          href="contributing.html">
           <span class="material-symbols-outlined text-base">volunteer_activism</span> Contributing
         </a>
         <a class="footer-link text-sm text-zinc-600 hover:text-primary dark:text-zinc-300 dark:hover:text-primary transition-all flex items-center gap-1.5"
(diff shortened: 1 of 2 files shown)


KumarNirupam1 (changes requested):
kindly address all the bot suggestions

TarunyaProgrammer (changes requested):
Thanks for the contribution. I found a few issues that should be addressed before this can be merged:

* `contributing.html` is referenced by both the footer and `vercel.json`, but the file is missing, so the new page currently results in a 404.
* The footer should link to the routed path (`/contributing`) instead of the relative path (`contributing.html`) to avoid incorrect resolution on nested routes.
* If this page is intended to work offline, `contributing.html` should also be added to the service worker's `CRITICAL_ASSETS` list.

Once these are fixed, the feature should work as intended.

KumarNirupam1 (changes requested):
thanks for the contribution! The idea of moving the contributing guide onto a dedicated page is a nice improvement, but I don't think this PR is ready to merge yet.

A few issues still need to be addressed:

The footer should link to the routed path (/contributing) instead of the relative path (contributing.html) to avoid incorrect resolution from nested routes.
Please ensure contributing.html is included in this PR. From the current diff, the page itself isn't present, which would result in a 404.
If the project's service worker pre-caches core pages, add the new contributing page to the pre-cache list so it also works offline.

Once these are addressed, the feature should be good to merge.

PR #2041, opened 2026-07-26
fix: auto-scroll focused organization card during keyboard navigation

### LINKED ISSUE #2038: Bug: Keyboard navigation does not scroll the focused organization card into view
**What is happening?**
When a user navigates organization cards with the arrow keys, the selected card is highlighted. But after moving beyond the visible cards, the page does not scroll to show the newly focused card.

**Why is this a problem?**
Keyboard users can lose track of which organization is selected. They may press Enter and open an organization that is not visible on screen.

**How to fix it:**

- Update the card-selection logic so scrollToFocused() can find the focused card. Either:
- add a data-filtered-idx attribute while rendering every card, or
- find the card by its position in the rendered card list instead of querying a missing attribute.
- 

**Steps to Reproduce:**

- Open the Organizations section.
- Use Arrow Down or Arrow Right repeatedly.
- Continue until focus moves past the initially visible cards.
- Notice that the focused card is not automatically scrolled into view.

**Files affected:**
src/js/app.js

Image

karizcloud:
## Summary

This PR fixes the keyboard navigation scrolling issue in the Organizations section.

### What Changed
- Updated the card selection logic so the focused organization card is automatically scrolled into view during keyboard navigation.
- Removed the dependency on the missing `data-filtered-idx` attribute by locating the focused card from the rendered card list.
- Preserved existing keyboard navigation, search, filtering, and selection behavior.

### Related Issue
Fixes #2038

### Type of Change
- [x] Bug fix

### Testing
- Verified keyboard navigation using Arrow Up/Down/Left/Right.
- Confirmed the focused card scrolls into view automatically.
- Verified search and filters continue to work correctly.
- Checked that there are no console errors.

### Checklist
- [x] Code follows the project style.
- [x] Tested locally.
- [x] No unrelated changes included.

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/2041?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -647,9 +647,15 @@ const GRID_COLS = () => {
 
 function scrollToFocused() {
   setTimeout(() => {
-    const g = document.getElementById('orgGrid');
-    const card = g?.querySelector(`[data-filtered-idx="${focusedIdx}"]`);
-    if (card) card.scrollIntoView({ block: 'nearest', behavior: 'smooth' });
+    const cards = document.querySelectorAll('#orgGrid article');
+    const card = cards[focusedIdx];
+    if (card) {
+      card.scrollIntoView({
+        behavior: 'smooth',
+        block: 'nearest',
+        inline: 'nearest'
+      });
+    }
   }, 30);
 }
 


S3DFX-CYBER (approved):
This fixes the keyboard-navigation scroll bug from #2038, where `scrollToFocused()` was silently failing because it looked up the focused card via `data-filtered-idx`, an attribute that was never being set anywhere in `app.js`. The new approach indexes directly into the rendered `<article>` elements via `document.querySelectorAll('#orgGrid article')[focusedIdx]`, which is simpler and actually works. `Fixes #2038` is linked, DCO is signed, the title is correctly lowercase `fix:`, and it's a tight 1-file +9/−3 diff with no dependency or lockfile changes. No security concerns — pure DOM queries, no `innerHTML` or untrusted data.

On the `index.html` inline / `src/js/app.js` sync rule: I checked, and `scrollToFocused` — along with the entire keyboard card-navigation feature (`focusedIdx`, `handleNavigationRight/Left/Down/Up`, `GRID_COLS`) — exists only in `src/js/app.js`, not in `index.html`'s inline script. So there's no sync violation here; the function being fixed only lives in one place. (That the feature is test-mirror-only and not in the production `index.html` runtime is a pre-existing architectural state, not something this PR introduces, so I'm not holding it against the change.)

The index alignment is sound: `updateCardFocus()` iterates the same `querySelectorAll('#orgGrid article')` NodeList with `idx === focusedIdx`, so `cards[focusedIdx]` in `scrollToFocused()` is consistent with how focus is tracked elsewhere. Adding `inline: 'nearest'` alongside `block: 'nearest'` is correct for the horizontal component of grid navigation and doesn't change vertical behavior.

Good to go.

PR #2053, opened 2026-08-03
fix(a11y): add keyboard navigation and ARIA labels to filter controls

### LINKED ISSUE #1969: [ACCESSIBILITY] Filter controls lack keyboard navigation and ARIA labels
### Description
The organization filter controls (dropdowns, checkboxes, search box) cannot be fully operated via keyboard alone, and lack proper ARIA labels. This makes the tool inaccessible to users who rely on keyboard navigation or screen readers.

### Steps to Reproduce
1. Navigate to the org finder
2. Press Tab to move focus through the page
3. Observe: filter dropdowns may not receive focus, or focus is not visible
4. Use a screen reader (VoiceOver/NVDA): the filter purpose is not announced

### Root Cause
Filter controls likely use custom `div`-based dropdowns without proper ARIA roles, or `select` elements without labels. Focus management may not be implemented for custom UI components.

### Impact
- Keyboard-only users (motor disabilities) cannot use the filtering system
- Screen reader users do not understand filter purposes or current values
- Violates WCAG 2.1 AA Success Criteria 1.3.1, 2.1.1, 4.1.2

### Proposed Fix
Add ARIA attributes and ensure keyboard operability:
```html
<!-- Filter dropdown -->
<label for="language-filter">Filter by Language</label>
<select id="language-filter" aria-label="Filter organizations by programming language">
  <option value="">All Languages</option>
  <!-- ... -->
</select>

<!-- Or for custom dropdown: -->
<div role="combobox" aria-expanded="false" aria-haspopup="listbox"
     aria-label="Filter by language" tabindex="0">
```
Also add visible focus indicators and keyboard event handlers for custom components.

I would like to audit and fix all accessibility issues in the filter UI if assigned.

waterWang:
## Summary

Fixes accessibility issues with the organization filter controls:
- Mobile filter and language dropdowns lacked proper ARIA attributes
- Dropdowns could not be closed with the Escape key
- Focus management was missing when dropdowns opened/closed
- Filter chips and pills lacked visible focus indicators
- Screen readers could not identify expanded/collapsed state of dropdowns

### HTML (index.html)
- **Mobile filter button**: Added `aria-expanded` and `aria-controls` attributes
- **Mobile filter panel**: Added `role="menu"` and `aria-hidden` for screen reader support
- **Mobile filter options**: Added `role="menuitem"` to each chip button
- **Mobile language button**: Added `aria-expanded` and `aria-controls` attributes
- **Mobile language panel**: Added `role="menu"` and `aria-hidden`
- Added `focus-visible:outline-*` Tailwind classes to all mobile filter buttons

### JavaScript (inline)
- **toggleMobileFilterDropdown()**: Now manages `aria-expanded` and `aria-hidden` for both filter and language panels
- **toggleMobileLangDropdown()**: Same ARIA management for language panel
- **Click-outside handler**: Updated to sync ARIA states when closing dropdowns
- **Escape key handler**: New! Closes open dropdown and returns focus to trigger button
- **Focus management**: Auto-focuses first item when dropdown opens

### CSS (inline stylesheet)
- Added `.filter-chip:focus-visible`, `.pill:focus-visible` styles with orange outline
- Added `.mobile-filter-option:focus-visible` style for mobile dropdown items
- Added `.clear-filters-btn:focus-visible` and `#loadMoreBtn:focus-visible` styles

## Testing
- ✅ All 25 existing tests pass
- ✅ Tested keyboard navigation: Tab through filter controls
- ✅ Escape key closes dropdowns and returns focus to trigger button
- ✅ ARIA attributes toggle correctly as dropdowns open/close

Closes #1969

<a href="https://cubic.dev/pr/S3DFX-CYBER/GSoC-Org-Finder-/pull/2053?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1174,6 +1174,29 @@ kbd:hover{
   }
 }
 
+/* Focus-visible styles for filter controls */
+.filter-chip:focus-visible,
+.pill:focus-visible {
+  outline: 2px solid var(--orange, #f46b0e);
+  outline-offset: 2px;
+  border-radius: 999px;
+}
+.mobile-filter-option:focus-visible {
+  outline: 2px solid var(--orange, #f46b0e);
+  outline-offset: 1px;
+  border-radius: 8px;
+}
+.clear-filters-btn:focus-visible {
+  outline: 2px solid var(--orange, #f46b0e);
+  outline-offset: 2px;
+  border-radius: 12px;
+}
+#loadMoreBtn:focus-visible {
+  outline: 2px solid var(--orange, #f46b0e);
+  outline-offset: 2px;
+  border-radius: 999px;
+}
+
 </style>
   <link rel="manifest" href="manifest.json" />
   <meta name="theme-color" content="#F46B0E" />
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1469,58 +1492,58 @@ <h2 class="text-4xl md:text-5xl font-extrabold font-headline tracking-tighter mt
...
         <span class="flex items-center gap-1.5">
           <span class="material-symbols-outlined text-sm text-primary">tune</span>
           <span id="mobileFilterLabel">Filters</span>
         </span>
         <span class="material-symbols-outlined text-sm transition-transform" id="mobileFilterChevron">expand_more</span>
       </button>
-      <div id="mobileFilterPanel" class="hidden absolute left-0 top-full mt-1 bg-white dark:bg-zinc-800 rounded-xl shadow-lg border border-zinc-100 dark:border-zinc-700 z-30 p-2 flex flex-col gap-1" style="min-width:180px">
-        <button id="mob-chip-veteran" data-mobile-chip="chip-veteran" onclick="toggleMobileChip(this)" class="mobile-filter-option flex items-center gap-2 px-3 py-2 rounded-lg text-xs font-medium hover:bg-surface-container-low transition-colors text-left">
+      <div id="mobileFilterPanel" role="menu" aria-hidden="true" class="hidden absolute left-0 top-full mt-1 bg-white dark:bg-zinc-800 rounded-xl shadow-lg border border-zinc-100 dark:border-zinc-700 z-30 p-2 flex flex-col gap-1" style="min-width:180px">
```
S3DFX-CYBER, line 1502: role="menu" / role="menuitem" without the full ARIA menu pattern (roving tabindex, arrow-key nav, aria-haspopup on trigger) creates a semantics/behavior mismatch. The language panel below has role="menu" but its child pills have no role="menuitem". Recommend dropping these roles and relying on the native button semantics + aria-expanded/aria-controls you already added.

index.html
```diff
@@ -1174,6 +1174,29 @@
...
+  outline: 2px solid var(--orange, #f46b0e);
+  outline-offset: 2px;
+  border-radius: 999px;
+}
+.mobile-filter-option:focus-visible {
+  outline: 2px solid var(--orange, #f46b0e);
+  outline-offset: 1px;
+  border-radius: 8px;
+}
+.clear-filters-btn:focus-visible {
```
S3DFX-CYBER, line 1189: This selector targets `.clear-filters-btn` but the element uses `id="clearFiltersBtn"` — this focus style will never apply. Use `#clearFiltersBtn:focus-visible` instead.

index.html
```diff
@@ -2674,43 +2697,102 @@ <h3 style="font-size:10px;font-weight:700;color:var(--muted);letter-spacing:.1em
...
+      langPanel.classList.add('hidden');
+      langPanel.setAttribute('aria-hidden', 'true');
+    }
+    if (langBtn) langBtn.setAttribute('aria-expanded', 'false');
     const lc = document.getElementById('mobileLangChevron');
     if (lc) lc.style.transform = '';
+    // Focus first item when opening
+    if (!isOpen) {
+      const firstItem = panel.querySelector('button, [tabindex]:not([tabindex="-1"])');
+      if (firstItem) setTimeout(() => firstItem.focus(), 50);
```
S3DFX-CYBER, line 2721: This setTimeout is never cancelled. If Escape or click-outside fires within 50ms, the callback still runs and focuses an element inside a hidden panel. Store the timer ID and clear it in every close path, or focus synchronously after toggling `hidden`.

S3DFX-CYBER (changes requested):
This adds keyboard navigation and ARIA attributes to the mobile filter and language dropdowns — Escape-to-close, focus management, aria-expanded/aria-controls on the trigger buttons, and focus-visible outlines on the chips and pills. The direction is solid and the intent is right, but there are a few things that need fixing before this can merge.

First, there's no linked issue in the PR body. The repo requires a `Closes #N` (or fix/resolve variant) — add one so the PR satisfies the hard rules and CI doesn't flag it.

The bigger problem is the ARIA semantics. The panels get `role="menu"` and the filter options get `role="menuitem"`, but the actual interaction doesn't implement the WAI-ARIA menu pattern — there's no roving tabindex, no arrow-key navigation, and the trigger has no `aria-haspopup="menu"`. Worse, the language panel gets `role="menu"` but its child language pills are left as plain `<button>` with no `role="menuitem"`, so that menu has no valid menuitem children at all. Screen readers will announce "menu" semantics that don't match how the control actually behaves, which undercuts the accessibility goal. The simpler and correct approach for a group of toggle filters is to drop the menu/menuitem roles entirely and rely on the native `<button>` names plus the `aria-expanded`/`aria-controls` you already added (which is correct). If you want a true menu, you'd need to implement the full pattern — but that's overkill here.

The focus management has a race condition. Both toggle functions call `setTimeout(() => firstItem.focus(), 50)` to focus the first item after opening, but that timeout is never tracked or cancelled. If the user presses Escape or clicks outside within that 50ms window — or opens the other dropdown — the pending timeout can still fire and move focus into a now-hidden panel, breaking the focus restoration the Escape handler tries to do. Focus the first item synchronously after toggling `hidden`, or store the timer ID and clear it in every close path (Escape, click-outside, sibling-open).

Finally, the `.clear-filters-btn:focus-visible` rule at the top of the diff targets a class that doesn't exist on the element — the Clear Filters button uses `id="clearFiltersBtn"`, not a `.clear-filters-btn` class. So that focus indicator will never apply. Target the existing ID instead.

One thing to verify: the `toggleMobileFilterDropdown` and `toggleMobileLangDropdown` functions live in the `index.html` inline script. If the same functions exist in `src/js/app.js` (the test mirror), they need to be updated there too — the repo's architecture requires the two stay in sync. If they don't exist in app.js, then this is fine as-is.

PR #2080, opened 2026-09-22
  fix: normalize "vue" to "vue.js" in skillExtractor

### LINKED ISSUE #2079: Bug: Skill normalization drops "Vue" → "vue.js" mapping, causing Vue-based orgs to be silently excluded from recommendations
### Description
`normalizeSkill()` in `src/js/skillExtractor.js` maps the dictionary token `"vuejs"` to `"vue.js"`, but has no mapping entry for the separate dictionary token `"vue"` — so a bare "vue" match is left as the literal string `"vue"` instead of being normalized to `"vue.js"`. Since org tags in `src/js/org.js` use the literal string `"vue.js"` (e.g. Learning Equality, omegaUp), and `normalizeSkill("vue.js")` maps to itself unchanged, a resume/GitHub profile skill extracted as `"vue"` never matches an org tag of `"vue.js"`. The two code paths (user skill extraction and org tag normalization) both call the same `normalizeSkill()` function, but because "vue" and "vue.js" are treated as distinct, unrelated skills, any Vue.js developer's language/topic score for Vue-tagged orgs is 0, and those orgs are pushed out of `getRecommendations()`'s top results (`src/js/recommender.js`) despite being a legitimate stack match.

### Steps to Reproduce
1. In a Node REPL/console, `require('./src/js/skillExtractor.js')` (or open the site and paste a resume) with text containing "Vue.js" (not "VueJS"/"vuejs").
2. Call `extractSkills("I have experience with Vue.js and Django")`.
3. Observe the returned skill is `"vue"`, not `"vue.js"`.
4. Feed this skill set into `getRecommendations()` and check organizations tagged `"vue.js"` (e.g. Learning Equality, omegaUp in `src/js/org.js`) — they receive 0 points from `calculateLanguageScore`/`calculateTopicScore` for the Vue match and rank far lower than they should.

### Expected Behavior
`normalizeSkill("vue")` should resolve to the same canonical value as `normalizeSkill("vuejs")` (i.e. `"vue.js"`), so a resume/profile mentioning "Vue" or "Vue.js" correctly matches organizations tagged `"vue.js"`.

### Actual Behavior
`extractSkills("...Vue.js...")` returns `"vue"`, while `extractSkills("...VueJS...")` returns `"vue.js"` — two different normalized values for what is effectively the same skill, so org tag matching (`orgTags.has(...)`) silently fails for the more common "Vue.js" spelling.

### Environment
- Browser: N/A (reproducible via pure JS logic, `src/js/skillExtractor.js`)
- OS: N/A
- Screen size: N/A

### Additional Context
Minimal fix: add `'vue': 'vue.js'` to the `mapping` object in `normalizeSkill()` (`src/js/skillExtractor.js`), consistent with how `'angular'` is already normalized to `'angularjs'`. Root cause is a one-line omission, but the impact is a silent, hard-to-notice ranking/matching defect in the core recommendation feature, so I'd rate this medium-to-high severity.

ayushi2577:
Closes #2079

## 📝 Description
`normalizeSkill()` in `src/js/skillExtractor.js` mapped the dictionary token
`"vuejs"` to `"vue.js"`, but had no mapping entry for the separate token
`"vue"` — so a bare `"vue"` match stayed as the literal string `"vue"`
instead of being normalized to `"vue.js"`.

Since org tags in `src/js/org.js` use the literal string `"vue.js"` (e.g.
Learning Equality, omegaUp), and both user-skill extraction and org-tag
normalization go through this same `normalizeSkill()` function, a resume or
GitHub profile mentioning "Vue.js" was never matching orgs tagged
`"vue.js"`. That silently zeroed the language/topic score for those orgs in
`calculateLanguageScore()` / `calculateTopicScore()`
(`src/js/recommender.js`), pushing genuinely good Vue matches out of the top
recommendations.

This PR adds the missing mapping so both spellings resolve to the same
canonical value, consistent with how `angular` is already normalized to
`angularjs`:

```js
'vue': 'vue.js',
```

<a href="https://cubic.dev/pr/S3DFX-CYBER/FindMyGSoC/pull/2080?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/js/skillExtractor.js
```diff
@@ -66,6 +66,7 @@ function normalizeSkill(skill) {
     'meteor.js': 'meteor.js',
     'angular': 'angularjs',
     'angularjs': 'angularjs',
+    'vue': 'vue.js',
```
S3DFX-CYBER, line 69: Kindly add assertions in tests/skills.test.js @ayushi2577

ayushi2577: i have added now do check and merge and lmk if any other changes are needed

